PR #1, opened 2025-01-09
Add base for Rust CI

connortsui20:
The basis for the CI here is taken from https://github.com/jonhoo/rust-ci-conf/. See `DOCS.md` for more information.

I added a temporary `optd-tmp` crate that we can remove once we figure out what needs to go there.

I've commented out the public semver checks as well as the code coverage part so we can easily turn those on later.

Note that this intentionally strict. Since we have a blank slate, we can amortize the dirty work needed to pass this CI.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+# Taken from https://github.com/jonhoo/rust-ci-conf/blob/main/.github/dependabot.yml
+
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
+    ignore:
+      - dependency-name: "*"
+        # patch and minor updates don't matter for libraries as consumers of this library build
+        # with their own lockfile, rather than the version specified in this library's lockfile
+        # remove this ignore rule if your package has binaries to ensure that the binaries are
+        # built with the exact set of dependencies and those are up to date.
+        update-types:
+          - "version-update:semver-patch"
+          - "version-update:semver-minor"
--- a/.github/DOCS.md
+++ b/.github/DOCS.md
@@ -1,0 +1,23 @@
+# Github config and workflows
+
+In this folder there is configuration for codecoverage, dependabot, and ci
+workflows that check the library more deeply than the default configurations.
+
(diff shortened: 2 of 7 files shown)


yliang412 (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
...
+    directory: /
+    schedule:
+      interval: daily
+    ignore:
+      - dependency-name: "*"
+        # patch and minor updates don't matter for libraries as consumers of this library build
+        # with their own lockfile, rather than the version specified in this library's lockfile
+        # remove this ignore rule if your package has binaries to ensure that the binaries are
+        # built with the exact set of dependencies and those are up to date.
+        update-types:
```
AlSchlo, line 19: Do we want to ignore minor and patches?
I believe we don't distribute optd as a library.

.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
...
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
```
AlSchlo, line 12: Daily is too much in the case we remove minor & patches.

.github/dependabot.yml
connortsui20: We don't distribute optd as a library now, but we likely will in the future (we want optd to be both a service and an embeddable library)

.github/dependabot.yml
connortsui20: The checks are daily but we only receive a notification if something is likely wrong

.github/dependabot.yml
AlSchlo: Yes, but it will start as a service. So we should get notified imo.

connortsui20: Actually I think it is the other way around right now, we will start as a library and then turn it into a service (we don't want to figure out our entire API immediately and keeping it as a library with a single process keeps it simple)

Also, this distinction has nothing to do with if we want to be aware of minor version updates or patches: even if we were to build a binary, we would only really want to know about specific _binary_ dependencies, not the library crate dependencies the binary crate itself would depend on. So that doesn't really make sense for this project.

PR #2, opened 2025-01-14
docs: add mdBook based documentation

yliang412:
## Problem

We need to record our design and retain knowledge of the codebase as a team. Documentation is one of the best ways to do it.

## Summary of changes

 - Check-in mdBook.
 - Add an initial overview that includes a wish list.
 - Add an initial glossary containing definitions and terms that optd developer will use.
 - Add a `rfcs/` directory with a template.
 - Add a pull request template (in `.github`).

## Future Work
- Setup gh-pages for the documentation. It might conflict with the benchmark site. We will delay deploying mdBook to gh-pages once that a decision is made


docs/README.md
```diff
@@ -0,0 +1,12 @@
+# Development Documentation
+
+The `src` folder contains the documentation on the optd query optimizer in the `mdBook` format.
+
+
+To view the documentation locally, you can follow the [`mdbook` installation guide](https://rust-lang.github.io/mdBook/guide/installation.html) to set up the environment. After installing `mdBook`, run the following command from the root of the optd repository:
```
connortsui20: Nit: `mdBook` has capital B

connortsui20 (approved):
Nit: Instead of pointing out every place, is it possible to remove all trailing whitespace and make sure there are newlines at the end of files?

This isn't really that important, it's just that I have my IDE set up to scream at me every time one of those things happens :smile:

docs/src/architecture/glossary.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Cascades
+
+### Expressions
+
+A **logical expression** is a tree/DAG of logical operators.
+
+A **physical expression** is a tree/DAG of physical operators.
+
+The term **expression** in the context of Cascades can refer to either a relational or a scalar expression.
```
SarveshOO7, line 20: Is there a difference between a relations expression and a relnode?

SarveshOO7: Is there a difference between scalar expressions and a scalarnode?

connortsui20: I've been thinking about "nodes" as the objects that we store in the memo table and "operators" as the actual trees/dag structures, but it might be good if we reversed this...

connortsui20 (changes requested):
On thinking about this a bit more after @SarveshOO7 pointed this out, I think it would be good to distinguish between nodes and operators, especially considering we need  to represent relations connected by pointers (in-memory) as well as connected via identifiers (on disk in the memo table).

Node: an in-memory node in a tree / DAG representing a relation / scalar
Operator: A representation of a relation / scalar that is not necessarily in memory

EDIT: OR

We could say everything that needs to be stored in the memo table is an expression (logical, physical, scalar expression), and then we say operators and nodes are purely in-memory (the initial input logical plan and final output physical plan are trees of operators / nodes)

I'm now in favor of the latter...

Could we maybe separate our the glossary from this PR?

docs/src/architecture/glossary.md
SarveshOO7: yes, I understand but the tree will be made of materialized nodes, right?

connortsui20:
See https://github.com/cmu-db/optd/blob/connor/types/types.md

docs/src/architecture/glossary.md
yliang412: Will fix in a seperate PR.

yliang412:
We can change the definitions. The main goal for this PR is to structure the docs directory.

PR #3, opened 2025-01-15
Add code coverage

connortsui20:
Requires a 85% coverage to pass. I'm not opposed to making this 80...

See [here](https://youtu.be/xUH-4y92jPg?si=_Yewt-jxI_S3BRmN&t=755) for explanation.

(TODO: set up everything else)

PR #4, opened 2025-01-15
Type Model

connortsui20:
This PR models almost all of the types that will be necessary for optimization. This includes:

- generic relational algebra operators that allow us to use the same "type" for both expressions in the memo table and operators in the plans
- logical / physical plans
- scalar operators and expressions
- partially materialized logical plans for rule binding
- transformation rule + implementation rule trait and some empty structs that implement them

I've named the crate itself `optd-core`. This can be subject to change, but I feel this is a reasonable default for now.

~~TODO: need to wait on #3 and #12 to be merged before proper CI checks can happen~~

Edit: I removed the `cargo rustdoc` check because its creating more problem than it would solve, see #14


AlSchlo (approved):
Renamed some variables for clarity, and moved around some code.

I added support to expressions and scalars.
Also added some MVP operators in there.

connortsui20:
@AlSchlo Several thoughts:

- Are we using the same `GroupId` for both relational and scalar groups? I actually don't have a strong preference either way, since that is easily changed in the future
- I don't really like the name `RelLink` because I still think that Rel is somewhat ambiguous (hence why I spelled out "relational" every time I used it). I also abandoned `Link` because I felt it was not descriptive enough
- Calling the root operator `node` instead of `root` I think is somewhat questionable, but I don't have any strong feelings about it
- I'm not totally convinced about the `LogicalExpression` and `PhysicalExpression` enums, as it implies that scalars can be both a logical expression and a physical expression. Based on what we talked about yesterday, this doesn't really make a lot of sense -- scalars should be a completely different type. So I'm inclined to keep a `RelationalExpression` and `ScalarExpression` type...

.github/workflows/check.yml
connortsui20, line 1: See #14

optd-core/src/operator/relational/physical/join/hash_join.rs
```diff
@@ -0,0 +1,12 @@
+/// Hash-based join operator that matches rows based on equality conditions.
+///
+/// Takes left and right input relations (`RelLink`) and joins their rows using
+/// a join condition (`ScalarLink`). Builds hash table from build side (right)
+/// and probes with rows from probe side (left).
+#[derive(Clone)]
+pub struct HashJoin<RelLink, ScalarLink> {
+    pub join_type: String,
+    pub probe_side: RelLink, // Left relation that probes hash table.
```
yliang412: nit

```suggestion
    /// Left relation that probes hash table.
    pub probe_side: RelLink,
```

optd-core/src/operator/relational/physical/join/hash_join.rs
```diff
@@ -0,0 +1,12 @@
...
+    pub build_side: RelLink, // Right relation used to build hash table.
```
yliang412: nit

```suggestion
    /// Right relation used to build hash table.
    pub build_side: RelLink,
```

AlSchlo:
> Are we using the same GroupId for both relational and scalar groups? 

Yes. We already have the differentiator in the "upper" layer of the enum, right?
> I don't really like the name RelLink because I still think that Rel is ambiguous

Me neither, but I liked child even less, as scalars are also children.
> Calling the root operator node instead of root 

Yes, since it's not always the root. I find that to be confusing.
> I'm not totally convinced about the LogicalExpression and PhysicalExpression enums

Requires some extra thought. Not too sure if we want Scalars as a completely different type. They are after all both logical and physical in Cascades.

optd-core/src/rules/implementation/hash_join.rs
```diff
@@ -0,0 +1,24 @@
+use super::*;
+use crate::operator::relational::{
+    logical::LogicalOperator,
+    physical::{join::hash_join::HashJoin, PhysicalOperator},
+};
+
+/// Implementation rule that converts a logical join into a hash join physical operator
+pub struct HashJoinRule;
+
+impl ImplementationRule for HashJoinRule {
```
yliang412, line 12: The `HashJoinRule` needs to check if the conditions satisfy the requirement.


```sql
-- check succeeds
t1 inner join t2 on t1.v1 = t2.v1

-- check succeeds
t1 inner join t2 on t1.v1 = t2.v1 and t1.v2 = t2.v2

-- check fails
t1 inner join t2 on t1.v1 > t2.v1
```
```suggestion
// TODO: rule may fail, need to check join condition
// https://github.com/cmu-db/optd/issues/15
impl ImplementationRule for HashJoinRule {
```

optd-core/src/operator/relational/physical/join/nl_join.rs
```diff
@@ -0,0 +1,11 @@
+/// Nested-loop join operator that matches rows based on a predicate.
+///
+/// Takes outer and inner relations (`RelLink`) and joins their rows using
+/// a join condition (`ScalarLink`). Scans inner relation for each outer row.
+#[derive(Clone)]
+pub struct NLJoin<RelLink, ScalarLink> {
+    pub join_type: String,
+    pub outer: RelLink, // Outer relation.
+    pub inner: RelLink, // Inner relation scanned for each outer row.
```
yliang412: nit: Same as `HashJoin`. Please use rust doc comments.

optd-core/src/operator/relational/physical/join/sort_merge_join.rs
```diff
@@ -0,0 +1,11 @@
+/// Sort-merge join operator that matches rows based on equality conditions.
+///
+/// Takes sorted left and right relations (`RelLink`) and joins their rows using
+/// a join condition (`ScalarLink`). Both inputs must be sorted on join keys.
+#[derive(Clone)]
+pub struct SortMergeJoin<RelLink, ScalarLink> {
+    pub join_type: String,
+    pub left_sorted: RelLink,  // Left sorted relation.
+    pub right_sorted: RelLink, // Right sorted relation.
```
yliang412: nit: Same as HashJoin. Please use rust doc comments.

optd-core/src/rules/transformation/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// filter pushdown under a `Join`).
+    ///
+    /// TODO: Ideally this should return a `Stream` instead of a fully materialized Vector.
+    async fn check_pattern(&self, expr: LogicalExpression, memo: &Memo) -> Vec<PartialLogicalPlan>;
+
+    /// Applies modifications to a partially materialized logical plan.
+    ///
+    /// These changes can create new logical or scalar expressions. However, note that
+    /// transformation rules will _not_ create new physical expressions.
+    fn apply(&self, expr: PartialLogicalPlan) -> Vec<LogicalExpression>;
```
yliang412: This interface does not look right. 

TransformationRule::apply might need to either:

1. take a `&mut Memo` to add a newly generated expression back from partial.

2. produces a vec of some kind of partial

optd-core/src/operator/relational/physical/join/sort_merge_join.rs
```diff
@@ -0,0 +1,11 @@
+/// Sort-merge join operator that matches rows based on equality conditions.
+///
+/// Takes sorted left and right relations (`Relation`) and joins their rows using
+/// a join condition (`Scalar`). Both inputs must be sorted on join keys.
+#[derive(Clone)]
+pub struct SortMergeJoin<Relation, Scalar> {
```
yliang412: nit: Let's call this a `MergeJoin`. We will use physical properties to ensure the inputs are sorted on the join keys. Sorting on the join keys will not be handled by the operator.

yliang412: Please change the enums and file/module names as well

optd-core/src/operator/scalar/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+/// - `Scalar`: Specifies whether the children scalars are other scalar operators or a group id.
+///
+/// This makes it possible to reuse the `ScalarOperator` type in [`LogicalPlan`],
+/// [`PhysicalPlan`], and [`PartialLogicalPlan`].
+///
+/// [`LogicalPlan`]: crate::plan::logical_plan::LogicalPlan
+/// [`PhysicalPlan`]: crate::plan::physical_plan::PhysicalPlan
+/// [`PartialLogicalPlan`]: crate::plan::partial_logical_plan::PartialLogicalPlan
+#[derive(Clone)]
+pub enum ScalarOperator<Scalar> {
```
yliang412, line 22: Regarding at which level we handle subquery, we can maybe have `ScalarOperator<Scalar, Relation>`, but only for the scalar operators that takes in a relational operator as children, you will have something like `Subquery<Relation, Scalar>`. For others, do what we have right now e.g. `Add<Scalar>`. 

We could do this in another PR.

yliang412 (commented):
Great work guys. This is solid foundation. Besides the nits, the two main improvements I want to point out us regarding the transformation rule API and ways we represent subqueries in the system.

For the transformation rule API, to add newly generated expressions back to the memo table  + recursively check for duplicates, you probably want to return some kind of partial plan.

For subqueries, I would argue we might want to handle it in some form of our representation. So there is the case where scalar could contain relation children. But after normalization, we should be fine. We might also want to support executing a subquery without unnesting (it will be a useful feature).

optd-core/src/rules/transformation/mod.rs
AlSchlo: I think it should produce a vec of partials. We should not insert anything in the memo here.

optd-core/src/operator/scalar/mod.rs
AlSchlo: Yes another PR it is.

yliang412: Created a follow-up issue here: https://github.com/cmu-db/optd/issues/16.

optd-core/src/operator/relational/logical/scan.rs
```diff
@@ -0,0 +1,9 @@
+/// Logical scan operator that reads from a base table.
+///
+/// Reads from table (`String`) and optionally filters rows using a pushdown predicate
+/// (`Scalar`).
+#[derive(Clone)]
+pub struct Scan<Scalar> {
+    pub table_name: String, // TODO(alexis): Mocked for now.
+    pub predicate: Option<Scalar>,
```
SarveshOO7, line 8: Based on our slack conversations, @yliang412 believes that have options here can result in complications when writing some rules. Perhaps, we should simply replace the None predicate type with a predicate type of True. (no predicate is the same as a always true predicate)

optd-core/src/expression.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::memo::GroupId;
+use crate::operator::relational::logical::LogicalOperator;
+use crate::operator::relational::physical::PhysicalOperator;
+use crate::operator::scalar::ScalarOperator;
+
+/// An expression in the memo table.
+///
+/// Can be either a logical expression (abstract operation) or a
+/// physical expression (specific implementation).
+pub enum Expression {
```
yliang412: Is this type helpful in cascades?

optd-core/src/operator/relational/physical/scan/table_scan.rs
```diff
@@ -0,0 +1,9 @@
+/// Table scan operator that reads rows from a base table
+///
+/// Reads from table (`String`) and optionally filters rows using
+/// a pushdown predicate (`Scalar`).
+#[derive(Clone)]
+pub struct TableScan<Scalar> {
+    pub table_name: String, // TODO(alexis): Mocked for now.
+    pub predicate: Option<Scalar>,
```
SarveshOO7, line 8: as I said in my other comment, we might want to remove the Option here. Simply replace a None predicate with a True predicate should work.

optd-core/src/rules/implementation/hash_join.rs
yliang412: follow-up issue created here: https://github.com/cmu-db/optd/issues/15. please add todo in code to reference the issue as well so we don't forget.

optd-core/src/plan/partial_logical_plan.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// A link to a scalar node in a [`PartialLogicalPlan`].
+///
+/// Can be either:
+/// - A materialized scalar operator node
+/// - A reference to an unmaterialized memo group
+#[derive(Clone)]
+pub enum Scalar {
+    Operator(Arc<ScalarOperator<Scalar>>),
+    Group(GroupId),
```
SarveshOO7: Nit: shouldn't this be GroupId(GroupId)
```suggestion
    GroupId(GroupId),
```

SarveshOO7 (changes requested):

optd-core/src/expression.rs
AlSchlo: Right now, I think not. We never need to group them.
Cascade only cares about Logical/Physical.

So I think we should not even have an independent Scalar type.

optd-core/src/operator/relational/logical/scan.rs
yliang412: We can probably keep it for now and then re-evaluate it later. I'm ok with adding the optional predicate if we only have it in scans. The thing I truly don't want is an optional predicate for all the relational operators.

AlSchlo: Good idea, but I will keep it as it for the time being. We shall see if it is really problematic in the future. Once we use these types a bit we'll quickly find out.

yliang412 (approved):

optd-core/src/rules/transformation/mod.rs
```diff
@@ -31,7 +31,7 @@ pub trait TransformationRule {
     ///
     /// These changes can create new logical or scalar expressions. However, note that
     /// transformation rules will _not_ create new physical expressions.
-    fn apply(&self, expr: PartialLogicalPlan) -> Vec<LogicalExpression>;
+    fn apply(&self, expr: PartialLogicalPlan) -> PartialLogicalPlan;
```
connortsui20, line 34: I still feel that there should be a way for the rule to tell the memo table what new expressions and groups needs to be added, it doesn't make much sense that the entire burden is on whoever inserts this back into the memo table

SarveshOO7 (approved):

optd-core/src/rules/transformation/mod.rs
yliang412: Example:

Same good old Filter push down predicates down below a Join:

Initial Plan:
```rust
(r) Filter            // group 10, scalar annotation: t2.v1 > 10 
    (s) Greater       // group 9
        (s) t1.v1     // group 8
        (s) 10        // group 7 
    (r) Join          // group 6, scalar annotation: t1.v1 = t2.v1
        (s) Eq        // group 5
            (s) t1.v1 // group 4
            (s) t2.v1 // group 3
        (r) Scan (t1) // group 1
        (r) Scan (t2) // group 2
```

match / "check_pattern":
```rust
(r) Filter            // group 10, scalar annotation: t2.v1 > 10 
    (s) Greater       // group 9
        (s) t1.v1     // group 8
        (s) 10        // group 7 
    (r) Join          // group 6, scalar annotation: t1.v1 = t2.v1
        (s) Eq        // group 5
            (s) t1.v1 // group 4
            (s) t2.v1 // group 3
        (r) GroupId(1)
        (r) GroupId(2)
```

After the apply:
```rust
(r) Join // NEW EXPR, OLD GROUP
   (s) Eq        // group 5, OLD
       (s) t1.v1 // group 4, OLD
       (s) t2.v1 // group 3, OLD
   (r) GroupId(1) // OLD
   (r) Filter // NEW EXPR, NEW GROUP
        (s) Greater       // group 9
            (s) t1.v1     // group 8
            (s) 10        // group 7 
```

If we return a partial plan, then we will have the information of which group `(r) Filter // NEW EXPR, NEW GROUP` belongs to before inserting the `Join // NEW EXPR, OLD GROUP`.

In your proposed solution, is there anything that deals with this?

You would not be able to construct the Join without inserting the filter first.
```rust
Join {
  condition: ScalarGroupId(5)
  left: RelationGroupId(1)
  right: ??????
}
```

connortsui20: > If we return a partial plan, then we will have the information of which group (r) Filter // NEW EXPR, NEW GROUP belongs to before inserting the Join // NEW EXPR, OLD GROUP.

This was my concern, at least the way I've defined a partial plan that information is not carried anywhere at all...

At the very least, returning the expressions themselves means we can distinguish between those new expressions and old ones

connortsui20: Hmm I think both of these are not fully complete interfaces... So I guess for now this is fine and we can change it later. Can we make an issue for this?

connortsui20: #17

optd-core/src/operator/relational/physical/join/merge_join.rs
```diff
@@ -0,0 +1,11 @@
+/// Merge join operator that matches rows based on equality conditions.
+///
+/// Takes sorted left and right relations (`Relation`) and joins their rows using
+/// a join condition (`Scalar`). Both inputs must be sorted on join keys.
+#[derive(Clone)]
+pub struct MergeJoin<Relation, Scalar> {
+    pub join_type: String,
+    pub left_sorted: Relation,  // Left sorted relation.
+    pub right_sorted: Relation, // Right sorted relation.
```
connortsui20: Rustdoc

optd-core/src/operator/relational/physical/join/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod hash_join;
+pub mod merge_join;
+pub mod nl_join;
```
connortsui20: I would prefer to see this spelled out

optd-core/src/operator/relational/physical/join/nl_join.rs
```diff
@@ -0,0 +1,13 @@
+/// Nested-loop join operator that matches rows based on a predicate.
+///
+/// Takes outer and inner relations (`Relation`) and joins their rows using
+/// a join condition (`Scalar`). Scans inner relation for each outer row.
+#[derive(Clone)]
+pub struct NLJoin<Relation, Scalar> {
```
connortsui20: I would prefer to see this spelled out

optd-core/src/operator/relational/physical/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+///
+/// [`PhysicalPlan`]: crate::plan::physical_plan::PhysicalPlan
+/// [`PhysicalExpression`]: crate::expression::PhysicalExpression
+#[derive(Clone)]
+pub enum PhysicalOperator<Relation, Scalar> {
+    TableScan(TableScan<Scalar>),
+    Filter(Filter<Relation, Scalar>),
+    Project(Project<Relation, Scalar>),
+    HashJoin(HashJoin<Relation, Scalar>),
+    NLJoin(NLJoin<Relation, Scalar>),
```
connortsui20: Same as above

optd-core/src/operator/scalar/constants.rs
```diff
@@ -0,0 +1,12 @@
+/// Constants that can appear in scalar expressions.
+#[derive(Clone)]
+pub enum Constant {
+    /// String constant (e.g. "hello")
```
connortsui20: Add periods

optd-core/src/plan/partial_logical_plan.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// A link to a scalar node in a [`PartialLogicalPlan`].
+///
+/// Can be either:
+/// - A materialized scalar operator node
+/// - A reference to an unmaterialized memo group
+#[derive(Clone)]
+pub enum Scalar {
+    Operator(Arc<ScalarOperator<Scalar>>),
+    GroupId(GroupId),
```
connortsui20, line 49: Needs to be `ScalarGroupId`

optd-core/src/operator/relational/physical/join/nl_join.rs
AlSchlo: I don't think this will ever be confusing

connortsui20: It's more about being consistent since we're spelling everything out now

optd-core/src/plan/partial_logical_plan.rs
AlSchlo: No need to right now.

PR #5, opened 2025-01-16
Update glossary and add lots of detail

connortsui20:
We have found that definitions in query optimization have internally become overloaded. This document defines key names and definitions for concepts that are required in optimization.

Don't look at the diff, it would be better to view this document directly: https://github.com/cmu-db/optd/blob/glossary/docs/src/architecture/glossary.md

TODO I don't really know how to define Properties and Rules well so I left the old explanation there.

### DIFF
--- a/docs/src/architecture/glossary.md
+++ b/docs/src/architecture/glossary.md
@@ -6,7 +6,7 @@ document defines key names and definitions for concepts that are required in opt
 Many of the names and definitions will be inspired by the Cascades framework. However, there are a
 few important differences that need to be addressed considering our memo table will be persistent.
 
-## Terms
+# Terms
 
 -   [Memo Table](#memo-table)
 -   [Group](#group)
@@ -20,7 +20,7 @@ few important differences that need to be addressed considering our memo table w
 -   [Plan](#logical--physical-plans)
     -   Logical Plan
     -   Physical Plan
--   [Operator / Plan Node](#logical--physical-plans)
+-   [Operator / Plan Node](#operators)
     -   Logical Operator
     -   Physical Operator
     -   Scalar Operator
@@ -32,7 +32,7 @@ few important differences that need to be addressed considering our memo table w
     -   Transformation Rule
     -   Implementation Rule
 
-## Comparison with Cascades
+# Comparison with Cascades
 
 In the Cascades framework, an expression is a tree of operators. In `optd`, we are instead defining
 a Plan to be a tree or DAG of operators. An Expression in `optd` strictly refers to the
@@ -42,17 +42,17 @@ See the [section below](#expression-logical-physical-scalar) for more informatio
 
 Most other terms in `optd` are similar to Cascades or self-explanatory.
 
-## Memo Table Terms
+# Memo Table Terms
 
 This section describes names and definitions of concepts related to the **Memo Table**.
 
-### Memo Table
+## Memo Table
 
 The **Memo Table** is the data structure used for dynamic programming in a top-down plan enumeration
 search algorithm. The memo table consists of a mutually recursive data structure made up of
 **Expressions** and **Groups**.
 
-### Expression (Logical, Physical, Scalar)
+## Expression (Logical, Physical, Scalar)
 
 An **Expression** is the representation of an operator **inside of the Memo Table**.
 
@@ -95,7 +95,9 @@ Another key difference between Plan Nodes and Expressions is that Expressions ha
 
 </details>
 
-### Expression Equivalence
+<br>
+
+## Expression Equivalence
 
 Two Logical Expressions are equivalent if the **Logical Properties** of the two Expressions are the
 same. In other words, the Logical Plans they represent produce the same set of rows and columns.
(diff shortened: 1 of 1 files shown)


docs/src/architecture/glossary.md
```diff
@@ -1,56 +1,135 @@
...
+## Memo Table Concepts
 
-A **scalar operator** (`ScalarNode`) describes an operation that can be evaluated to obtain a single value. In other literature this is also referred to as a sql expression or a row expression.
+This section describes names and definitions of concepts related to the **Memo Table**.
 
-## Cascades
+### Memo Table
 
-### Expressions
+The **Memo Table** is the data structure used for dynamic programming in a top-down plan enumeration
```
yliang412:
```suggestion
The **memo table** is the data structure used for dynamic programming in a top-down plan enumeration
```

nit: I feel bold is enough, no need to Capitalize all the terms.

docs/src/architecture/glossary.md
```diff
@@ -1,56 +1,135 @@
...
-### Relational operator
-A **relation operator** (`RelNode`) describes an operation that can be evaluated to obtain a bag of tuples. In other literature this is also referred to as a query plan. A relational operator can be either logical or physical.
+Many of the names and definitions will be inspired by the Cascades framework. However, there are a
+few important differences that need to be addressed considering our memo table will be persistent.
 
-### Scalar operator
+## Memo Table Concepts
 
-A **scalar operator** (`ScalarNode`) describes an operation that can be evaluated to obtain a single value. In other literature this is also referred to as a sql expression or a row expression.
+This section describes names and definitions of concepts related to the **Memo Table**.
```
yliang412:
```suggestion
This section describes names and definitions of concepts related to the [**memo table**](#memo-table).
```

It might be good to add links to the relevant sections when possible.

yliang412 (approved):
Overall looks great. I posted comments related to links and styles, but I think you already started to add a lot of those.

docs/src/architecture/glossary.md
connortsui20: Yeah I'll do links instead of capitalization

docs/src/architecture/glossary.md
```diff
@@ -1,58 +1,185 @@
...
+engine to produce a table. This can also be referred to as a Physical Query Plan. The Operators that
+make up this Physical Plan can be considered Physical Plan Nodes.
+
+## Operators
+
+A **Logical Operator** is a node in a Logical Plan (which is a tree or DAG).
+
+A **Physical Operator** is a node in a Physical Plan (which is a tree or DAG).
+
+A **Scalar Operator** describes an operation that can be evaluated to obtain a single value. This
```
SarveshOO7: I will start this conversation here.
Examples of physical operators: Hash Join, SMJ, Index Scan, etc.
Examples of logical operators: Scan, Join, etc.
Example of physical plan nodes:  HashJoin(A, B), IndexScan(C), etc.
Example of logical plan nodes: Join(A, B), Scan(C), etc.

Example of scalar operator: >, <, <=, >=, ==, etc.
Example of scalar plan nodesL: A>B, C== 3, etc.

Do these sound right to you @connortsui20 ?

connortsui20: I'm thinking about just having a section for every single term instead of sprinkling it around, but that means this document is going to budge massive...

SarveshOO7: That is fine with me. But do these sound right to you?

connortsui20: I'm not so sure about the scalar stuff (@yliang412 could you take a look) but I also have something similar in the drop-down details for Expression section above so yes

SarveshOO7: My last question is that in the example for physical plan node, HashJoin(A,B)

Are A and B physical plan nodes? Can they be logical plan nodes based on what point we are at during the query evaluation?

docs/src/architecture/glossary.md
```diff
@@ -32,70 +33,106 @@ few important differences that need to be addressed considering our memo table w
...
-The **Memo Table** is the data structure used for dynamic programming in a top-down plan enumeration
+The memo table is the data structure used for dynamic programming in a top-down plan enumeration
 search algorithm. The memo table consists of a mutually recursive data structure made up of
-**Expressions** and **Groups**.
+[Expression]s and [Group]s.
 
 ## Expression (Logical, Physical, Scalar)
 
-An **Expression** is the representation of an operator **inside of the Memo Table**.
+An **expression** is the representation of a non-materialized operator _inside_ of the [Memo Table].
```
SarveshOO7: This is perfect. it is much more clearer

SarveshOO7 (approved):
The definitions look good to me

connortsui20:
I'm making more changes (there are still todos everywhere) so please don't merge.

Also, @yliang412 there is a fixme that I don't fully understand, could you try rewriting that sentence if you have the time?

docs/src/architecture/glossary.md
connortsui20: This is a TODO now, and since there are so many TODOs I'm going to resolve this and leave this to figure out during a sync on Tuesday

PR #8, opened 2025-01-19
Add details to rfcs template

connortsui20:
## Problem

The RFCs template was sort of bare, so I wanted to make it more detailed. A lot of stuff is copied from the Rust RFCs template [here](https://github.com/rust-lang/rfcs/blob/master/0000-template.md)

## Summary of changes

Changed some of the sections and added details on how to write each one.


yliang412 (approved):
Thanks. I like the idea of adding more details to the RFC template. It also helps people think about their designs.

SarveshOO7 (approved):
LGTM

PR #10, opened 2025-01-20
WIP Built some initial e2e datafusion infrastructure

SarveshOO7:
## Problem
We want to be able to run our queries on datafusion.

## Summary of changes
The only update I have is that I have setup the datafusion e2e to be able to run multiple queries and print the results and the plans. Currently, since we do not have an optimizer built out and we don’t have an logical plan IR to convert the datafusion logical plan into, I am simply running the datafusion optimizer.


connortsui20 (changes requested):
Should we call this optd-datafusion instead or something more descriptive than "infra"?

Make sure to fix all of the clippy lints (you can either look at the `Files changed` tab on GitHub or just run `cargo clippy --all-targets`

Also, please add comments on everything because I can't really understand the code...

See the official [Rust style guide](https://doc.rust-lang.org/nightly/style-guide/) as well for things that clippy cant check (things like full punctuation on comments)

connortsui20:
It might be a good idea to fully rebase everything on top of main, something like `git rebase -i $(git merge-base HEAD master)` and just squash everything into a single commit

SarveshOO7:
Yes, I agree. Most of what was done here has become useless. However, I think I'll still try and make a version of this stuff work on this branch and then worry about moving stuff around later.

yliang412:
Integrated into https://github.com/cmu-db/optd/pull/26

PR #11, opened 2025-01-24
feat: minimalistic storage layer setup

yliang412:
## Problem

Persistence storage is on top of the optd optimizer wishlist.

## Summary of changes

-  models the database schema and brings it to life using [Diesel](https://diesel.rs/guides/) as the ORM.
- The schema uses an operator per table model.
- implemented a storage manager API on top of the ORM, currently it support:

```rust
impl StorageManager {
     pub fn add_logical_expr(&mut self, logical_expr: LogicalExpr) -> (LogicalExprId, RelGroupId);
    
     pub fn add_logical_expr_to_group(
            &mut self,
            logical_expr: LogicalExpr,
            rel_group_id: RelGroupId,
        ) -> LogicalExprId;
    
    pub fn get_logical_expr_identifiers(
            &mut self,
            logical_expr: &LogicalExpr,
        ) -> Option<(LogicalExprId, RelGroupId)>;
    
    
    pub fn get_all_logical_exprs_in_group(
            &mut self,
            rel_group_id: RelGroupId,
        ) -> Vec<LogicalExprWithId>;

    // more private methods ...
}
```

## Future works

This PR unlocks the opportunity for people to work on rule matching. A fully sketched-out schema will be implemented gradually this week and next week.


yliang412:
## Demo

Imagine the following schema and query:

```sql
CREATE TABLE t1(v1 INTEGER, v2 TEXT);
CREATE TABLE t2(v1 INTEGER, v2 TEXT);
SELECT * from t1 inner join t2 on t1.v1 = t2.v1 where t1.v2 = 'foo';
- LogicalFilter (on: t1.v2 = 'foo')
  - child: LogicalJoin (inner, on t1.v1 = t2.v1)
    - left: LogicalScan (t1)
    - right: LogicalScan (t2)
```

Besides adding it in initially, we also apply the join commutativity rule and register the equivalent of

```
- LogicalJoin (inner, on t1.v1 = t2.v1)
  - left: LogicalScan (t2)
  - right: LogicalScan (t1)
```

```shell
$ cargo run --bin storage_demo_with_trait
```


After running the demo, you will have:

```shell
$ sqlite3 test_memo.db
```

```sql
sqlite> select l.id, l.group_id, l.created_at, desc.name from logical_exprs as l, logical_op_kinds as desc where l.logical_op_kind_id = desc.id;
┌────┬──────────┬─────────────────────┬───────────────┐
│ id │ group_id │     created_at      │     name      │
├────┼──────────┼─────────────────────┼───────────────┤
│ 1  │ 1        │ 2025-01-24 05:52:45 │ LogicalScan   │
│ 2  │ 2        │ 2025-01-24 05:52:45 │ LogicalScan   │
│ 3  │ 3        │ 2025-01-24 05:52:45 │ LogicalJoin   │
│ 4  │ 3        │ 2025-01-24 05:52:45 │ LogicalJoin   │
│ 5  │ 4        │ 2025-01-24 05:52:45 │ LogicalFilter │
└────┴──────────┴─────────────────────┴───────────────┘

sqlite> select * from logical_joins;
┌─────────────────┬───────────┬──────┬───────┬───────────────┐
│ logical_expr_id │ join_type │ left │ right │   join_cond   │
├─────────────────┼───────────┼──────┼───────┼───────────────┤
│ 3               │ 0         │ 1    │ 2     │ t1.v1 = t2.v1 │
│ 4               │ 0         │ 2    │ 1     │ t1.v1 = t2.v1 │
└─────────────────┴───────────┴──────┴───────┴───────────────┘
sqlite> select * from logical_scans;
┌─────────────────┬────────────┐
│ logical_expr_id │ table_name │
├─────────────────┼────────────┤
│ 1               │ t1         │
│ 2               │ t2         │
└─────────────────┴────────────┘
sqlite> select * from logical_filters;
┌─────────────────┬───────┬───────────────┐
│ logical_expr_id │ child │   predicate   │
├─────────────────┼───────┼───────────────┤
│ 5               │ 3     │ t1.v2 = 'foo' │
└─────────────────┴───────┴───────────────┘
```

It is also fine if you run them multiple times. The created_at timestamp won't change.

optd/src/storage/models/common.rs
```diff
@@ -0,0 +1,122 @@
...
+    RightAnti,
+    LeftMark,
+}
+
+impl TryFrom<i32> for JoinType {
+    type Error = anyhow::Error;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        use JoinType::*;
+        match value {
```
skyzh, line 83: consider using a 3rd party crate to reduce boilerplate code

optd/src/storage.rs
```diff
@@ -0,0 +1,140 @@
...
+            .select((
+                LogicalExprRecord::as_select(),
+                schema::logical_op_kinds::dsl::name,
+            ))
+            .load::<(LogicalExprRecord, String)>(&mut self.conn)
+            .unwrap();
+
+        let mut exprs = Vec::with_capacity(records.len());
+
+        for (record, name) in records {
```
skyzh, line 97: sounds like a lot of round trips to the database

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+#[enum_dispatch(LogicalExprStorage)]
+#[derive(Debug)]
+pub enum LogicalExpr {
+    Scan(LogicalScan),
+    Filter(LogicalFilter),
+    Join(LogicalJoin),
+}
+
+#[derive(Debug)]
+pub struct LogicalExprWithId {
```
skyzh, line 42: just FYI the old codebase doesn't have this and I'm unsure how useful it is to have it

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+pub trait LogicalExprStorage {
+    fn name(&self) -> &'static str;
+
+    /// Gets the logical expression id if it is already in the database.
+    fn id(&self, storage: &mut StorageManager) -> Option<LogicalExprId>;
+
+    fn insert_op(&self, id: LogicalExprId, storage: &mut StorageManager);
+
+    /// Gets the logical operator kind id.
+    fn op_kind(&self, storage: &mut StorageManager) -> LogicalOpKindId {
```
skyzh, line 62: do we have plan to use the diesel async interface? sounds like we are starting with sync first and needs to do a significant refactor

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+        let rel_group_id = storage.rel_group_of_logical_expr(id);
+        Some((id, rel_group_id))
+    }
+
+    /// Adds the logical expression to the database.
+    /// If the logical expression is already in the database, it returns the logical expression id and the relational group id.
+    /// Otherwise, it inserts the logical expression into the database and returns the generated logical expression id and
+    /// the relational group id.
+    fn add(&self, storage: &mut StorageManager) -> (LogicalExprId, RelGroupId) {
+        if let Some((id, rel_group_id)) = self.get_identifiers(storage) {
```
skyzh, line 84: do check and insert in one sql using `insert on conflict`?

skyzh: TODO: wrap all these ops into a txn

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[enum_dispatch]
+pub trait LogicalExprStorage {
+    fn name(&self) -> &'static str;
+
+    /// Gets the logical expression id if it is already in the database.
+    fn id(&self, storage: &mut StorageManager) -> Option<LogicalExprId>;
+
+    fn insert_op(&self, id: LogicalExprId, storage: &mut StorageManager);
```
skyzh, line 59: what does this function do and any example using it?

skyzh (approved):
Generally LGTM with the approach -- I could see some minor problems and we can figure out them later.

skyzh:
I also recommend having a single counter for both group id and expr id so that one number uniquely identifies an entity in the system. This helps with debugging. i.e., if we assign 1 as a group id, then 1 won't be used as an expression id. It is possible with Postgres's `create sequence`. (Something for the future)

optd/src/storage/models/logical_expr.rs
yliang412: I'm looking into how diesel-async works. Would there be a lot of problems besides adding in the await points?

optd/src/storage/models/logical_expr.rs
yliang412: > do check and insert in one sql using `insert on conflict`?

Yea, I think to do this we probably need to build an index on the data columns and then you can start using upserts.

optd/src/storage/models/logical_operators.rs
```diff
@@ -0,0 +1,30 @@
+//! The logical operator objects in the optimizer storage layer.
```
connortsui20, line 1: This should be a `mod.rs` file considering that we're going to have tens of operators and the directory structure is going to super difficult to navigate if the submodule root is located in a different place than the operator sub-submodules

optd/src/storage/models/physical_operators.rs
```diff
@@ -0,0 +1,30 @@
+//! The physical operator objects in the optimizer storage layer.
```
connortsui20, line 1: This should also be a `mod.rs` file

optd/src/storage/models.rs
```diff
@@ -0,0 +1,6 @@
+pub mod common;
+pub mod logical_expr;
+pub mod logical_operators;
+pub mod physical_expr;
+pub mod physical_operators;
+pub mod rel_group;
```
connortsui20, line 6: This definitely needs to be in a `mod.rs` file

optd/src/storage/schema.rs
```diff
@@ -0,0 +1,116 @@
+// @generated automatically by Diesel CLI.
```
connortsui20, line 1: So I know that this is code generated, but is it possible to put comments here anyways? The workflow that worked for the codegen from sea-orm was that we added extra stuff into this file (like comments and lints), and when the codegen overwrote the extra stuff we just used git to put them back

connortsui20 (changes requested):
First pass (super high level) review:

I added some comments about module file structure

In terms of style, I think we should be very intentional about naming. Things like `Expr` are pretty self explanatory, but I actually have to think hard when stumbling across something like `physical_op_kinds` and `rel_group` -- it will be even harder for someone new jumping into this codebase. Default to spelling out everything unless it becomes unwieldy (like in the case of writing out the entire `Expression`). Things like `Rel` and `OpKind` are, in my opinion, hard to grok.

.github/workflows/check.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         uses: dtolnay/install@cargo-docs-rs
       - name: cargo docs-rs
         # TODO: Once we figure out the crates, rename this.
-        run: cargo docs-rs -p optd-tmp
+        run: cargo docs-rs -p optd
```
connortsui20, line 94: I still personally think that this should be called `optd-storage` instead of just `optd` because it will become confusing

.gitignore
```diff
@@ -18,4 +18,11 @@ Cargo.lock
...
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
\ No newline at end of file
+#.idea/
+
+### Project Specific ###
+
+# The memo table database for testing purposes.
+test_memo.db
```
connortsui20: I think `*.db` makes more sense here

Cargo.toml
```diff
@@ -1,3 +1,17 @@
 [workspace]
-members = ["optd-tmp"]
+members = ["optd"]
 resolver = "2"
+
+[workspace.dependencies]
+anyhow = "1"
```
connortsui20, line 6: We probably want to use [`snafu`](https://docs.rs/snafu/latest/snafu/) given the service-like-library nature of this project

optd/migrations/.keep
connortsui20, line 1: Is this important?

connortsui20 (changes requested):

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+
+// Identifier for a logical expression.
+define_diesel_new_id_type_from_to_sql!(LogicalExprId, i64, BigInt);
+
+#[derive(Debug, Queryable, Selectable, Identifiable, AsChangeset)]
+#[diesel(table_name = logical_exprs)]
+#[diesel(belongs_to(RelGroup))]
+#[diesel(belongs_to(LogicalOpKind))]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct LogicalExprRecord {
```
connortsui20, line 23: I don't think we ever talked about this type, this seems to be a new table? I see that it is used in one other place in `get_all_logical_exprs_in_group`, but I don't immediately see the purpose of this...

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub id: LogicalExprId,
+    /// The type descriptor of the logical expression.
+    pub logical_op_kind_id: LogicalOpKindId,
+    /// The relational group that this logical expression belongs to.
+    pub group_id: RelGroupId,
+    /// The time at which this logical expression was created.
+    pub created_at: chrono::NaiveDateTime,
+}
+
+#[enum_dispatch(LogicalExprStorage)]
```
connortsui20: For now we can use `enum_dispatch`, but in the future we will likely want to write out the entire thing because it gives us more control of what is happening (and better error messages).

optd/src/storage/models/logical_expr.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct LogicalExprWithId {
+    pub id: LogicalExprId,
+    pub inner: LogicalExpr,
+}
+
+pub trait LogicalOp {
+    const NAME: &'static str;
+
+    fn get(id: LogicalExprId, storage: &mut StorageManager) -> Self;
+}
```
connortsui20: I think this should be called `LogicalOperator`, and I also think it should be defined somewhere else

optd/src/storage.rs
connortsui20, line 1: I think there should be a `storage_manager` submodule in the `storage` module that only contains the `StorageManager` type and impl

connortsui20 (changes requested):

optd/src/storage/models/logical_expr.rs
connortsui20: actually my bad, I think we should avoid using `enum_dispatch` in general and manually implement shared behavior between types by wrapping them in an enum ourselves. Basically, I think we should stay away from traits in general for now, especially traits that are not object safe.

optd/migrations/.keep
yliang412: This is generated. I did not touch it.

optd/src/storage/models/logical_expr.rs
yliang412: This is the flat record type on the logical_exprs table. You need this and the specific operator table to get out the full `LogicalExpr` enum. Subject to change but I would argue this does not change to overall API.

.github/workflows/check.yml
yliang412: I think last time we decided initially we are going to have a single crate called `optd`.  `storage` will be a module inside that crate.

optd/src/storage/models/logical_operators.rs
yliang412: I prefer having a consistent project structure and avoid using mod. With the same convention, people will know where to find submodules.

connortsui20: Yes I agree, and I'm saying that there is not much justification here in having a `logical_operator.rs` file located so far away from the `logical_operator` folder, especially since we're going to have a lot of modules.

See discussions:
- https://www.reddit.com/r/rust/comments/18pytwt/noob_question_foomodrs_vs_foors_foo_for_module/
- https://users.rust-lang.org/t/module-mod-rs-or-module-rs/122653
- https://internals.rust-lang.org/t/the-module-scheme-module-rs-file-module-folder-instead-of-just-module-mod-rs-introduced-by-the-2018-edition-maybe-a-little-bit-more-confusing/21977/17?u=zirconium-n

This is obviously not a black and white conversation, but in my opinion the only argument against having `mod.rs` is that you have multiple files named the same thing. But with the last link listed above plus the fact that `mod.rs` should really only be re-exporting stuff, I don't see any good reason to follow this.

optd/migrations/.keep
connortsui20: https://github.com/diesel-rs/diesel/issues/1399#issuecomment-1442360648

I think we should remove it

optd/src/storage/models/logical_expr.rs
connortsui20: If that's the case, can you make the visibility `pub(super)` or whatever is the lowest visibility for the memo table to have access to it?

optd/src/storage/models/logical_expr.rs
yliang412: Removed enum dispatch in [037ded3](https://github.com/cmu-db/optd/pull/11/commits/037ded399a2233814f2a089032b4f91bf467c453)

optd/src/storage/models/logical_expr.rs
yliang412: See [037ded3](https://github.com/cmu-db/optd/pull/11/commits/037ded399a2233814f2a089032b4f91bf467c453), removed enum_dispatch. Do feel more boilerplate to write. I might reconsider it later.

.gitignore
yliang412: [32857c2](https://github.com/cmu-db/optd/pull/11/commits/32857c2a405e1c7f18bee4e61b1b2dbbaf460aac)

optd/src/storage/models/logical_expr.rs
yliang412: This is used by `add` to add an entry for per-operator tables. This reduces the LoC needed for each operator.

optd/src/storage/models/logical_expr.rs
yliang412: I would also think that an in-memory caching layer is necessary to speed things up.

.github/workflows/check.yml
connortsui20: If that's the case, then we arguably shouldn't have a workspace at all.

With what @SarveshOO7 was working on, I think it makes sense to have a separate crate for everything around the pipeline we talked about because we _do not_ want to have to recompile datafusion every single time we make a change.

I think either we completely remove the workspace and call it `optd`, or we start with 2 crates (something like `optd-core` and `optd-datafusion`) and make sure we don't expand beyond 2 without good reason.

optd/migrations/2025-01-22-223443_create_physical_op_kinds/up.sql
```diff
@@ -0,0 +1,8 @@
+-- The physical operator descriptor table stores all the 
+-- physical operators that can be used in the optimizer.
+CREATE TABLE physical_op_kinds (
```
AlSchlo, line 3: TBH, I think having an enum here might be more practical. It will also make all the queries simpler and safe us a join. Might as well codegen everywhere, right?

This table will never change at run time.

AlSchlo: I'll experiment with that during the WE.

optd/src/bin/storage_demo_manual.rs
```diff
@@ -0,0 +1,271 @@
...
+        .filter(logical_op_kinds::name.eq("LogicalScan"))
+        .select(logical_op_kinds::id)
+        .first::<LogicalOpKindId>(&mut storage.conn)?;
+
+    let logical_join_id = logical_op_kinds::table
+        .filter(logical_op_kinds::name.eq("LogicalJoin"))
+        .select(logical_op_kinds::id)
+        .first::<LogicalOpKindId>(&mut storage.conn)?;
+
+    let logical_filter_id = logical_op_kinds::table
```
AlSchlo, line 30: if we keep the table approach, I guess we could have a big function that loads all of them & caches the ids in the future.

optd/migrations/2025-01-22-223443_create_physical_op_kinds/up.sql
connortsui20: I agree, even though I haven't actually given the logic of this PR a review, we should be pushing as much work as we can to compile-time / codegen / preprocessing, _especially_ since we want the foundation to be minimal at runtime so that we don't run into a wave of runtime bugs in the future

optd/src/bin/storage_demo_manual.rs
yliang412: We could also always carry the operator kind.

optd/migrations/2025-01-22-223940_create_logical_exprs/up.sql
```diff
@@ -0,0 +1,14 @@
+-- The relational logical expressions table specifies 
+-- which group a logical expression belongs to.
+CREATE TABLE logical_exprs (
+    -- The logical expression id.
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
```
connortsui20, line 5: If we go with Chi's suggestion of having a unique ID for each object we shouldn't use autoincrement here

optd/migrations/2025-01-22-223940_create_logical_exprs/up.sql
```diff
@@ -0,0 +1,14 @@
...
+    -- The logical operator descriptor id.
+    logical_op_kind_id BIGINT NOT NULL,
```
connortsui20, line 7: Probably doesn't need to be `BIGINT` here? Or do foreign key references HAVE to be `BIGINT`?

optd/migrations/2025-01-22-223441_create_relational_groups/up.sql
```diff
@@ -0,0 +1,8 @@
+-- A relational group contains a set of relational expressions 
+-- that are logically equivalent.
+CREATE TABLE rel_groups (
```
connortsui20, line 3: Yeah I think expand out the "relational" here, and also be consistent with what the migration file directory is called too

optd/migrations/2025-01-22-223441_create_relational_groups/up.sql
```diff
@@ -0,0 +1,8 @@
...
+    -- The group identifier.
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    -- Time at which the group is created.
+    created_at TIMESTAMP DEFAULT (CURRENT_TIMESTAMP) NOT NULL
+);
```
connortsui20, line 8: Add TODOs here saying that we will add other columns like an optional winner, potentially cost, and also other group metadata related to our union-find parent pointer idea

optd/migrations/2025-01-22-223940_create_logical_exprs/up.sql
connortsui20, line 1: It seems like there might be an unnecessary level of indirection here?

For the sake of this example, suppose we only have logical expression types. My understanding of this is that there is a `logical_exprs` table that tracks expression IDs and maps them to their logical operator kind (Scan, Filter, Join). The purpose of this is so that every logical expression ID lookup can be paired with a logical operator kind so that we know what specific table to go look at.

What if instead of having a full table that makes this mapping, we instead encode the operator kind _inside_ of the expression ID? For example, we could have it such that the upper 16 bits of a 64-bit ID encode the operator kind, and the lower 48 bits encode the unique expression given the operator kind (we would likely not even need all 16 bits, we could probably get away with the top 8 bits even).

You could probably argue that this introduces complexity, but in my mind I feel that having extra tables that are not strictly necessary introduces even more complexity. When I was trying to read this PR I had to spend a non-trivial amount of time trying to understand the architecture, since I did not expect that specific table when we initially talked about the architecture in previous meetings.

In terms of implementation of the above proposal, yes it would be a bit ugly, but if abstracted correctly at the storage layer level, there should only really be 1 function that handles the conversion into some sort of strongly-typed struct like this:

```rust
struct ExprId {
    kind: u16, // could also encode logical/physical difference before converting into stronger types
    id: u64,   // truncated at 48 bits
}
```

This is opposed to having a dedicated table that makes this mapping, where every single expression lookup now has to look up and extra record on disk (and even in memory, this roundtrip is going to need to happen for every single lookup during optimization).

As for the group ID mapping, I think that needs to be in a separate junction table regardless for group operations. Unless this is actually supposed to be the junction table? If that's the case, then all 3 things in this table would have needed to be a composite primary key tuple.

I think there are tradeoffs to both models, but I am leaning towards removing the layer of indirection.

connortsui20 (changes requested):

PR #18, opened 2025-01-30
First PR cleanup

connortsui20:
## Problem

#4 Was the first major PR, but since we were making a lot of changes last minute there were a few things that slipped through with respect to quality control.

## Summary of changes

- Added very strict clippy documentation requirements. Discussion below.
- Added documentation for everything except the `operator::relational::physical` module and the `operator::scalar` module, as those seem kind of unstable right now.
- Changed the scalar generic param of expressions to use `ScalarGroupId` instead of `GroupId`.
- Added several TODOs that I think will be resolved pretty quickly, though I'd appreciate it if others took a look at those in the case that we can actually get those done now.

Now that we have something to work off of, requiring documentation on everything (including private items) means that we'll pay the cost right now of making sure people understand what we're doing in the future. I'm willing to relax it a little bit if we put _other_ stuff in place that ensure nobody in the future encounters large chunks of code with zero documentation.

### DIFF
--- a/optd-core/src/operator/relational/logical/filter.rs
+++ b/optd-core/src/operator/relational/logical/filter.rs
@@ -8,5 +8,8 @@ pub struct Filter<Relation, Scalar> {
     /// The input relation.
     pub child: Relation,
     /// The filter expression denoting the predicate condition for this filter operation.
+    ///
+    /// For example, a filter predicate could be `column_a > 42`, or it could be something like
+    /// `column_b < 100 AND column_c > 1000`.
     pub predicate: Scalar,
 }
--- a/optd-core/src/operator/relational/logical/filter.rs
+++ b/optd-core/src/operator/relational/logical/filter.rs
@@ -1,8 +1,12 @@
+//! A logical filter.
+
 /// Logical filter operator that selects rows matching a condition.
 ///
 /// Takes input relation (`Relation`) and filters rows using a boolean predicate (`Scalar`).
 #[derive(Clone)]
 pub struct Filter<Relation, Scalar> {
+    /// The input relation.
     pub child: Relation,
+    /// The filter expression denoting the predicate condition for this filter operation.
     pub predicate: Scalar,
 }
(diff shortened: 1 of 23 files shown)


optd-core/src/operator/relational/logical/filter.rs
```diff
@@ -1,8 +1,12 @@
+//! A logical filter.
+
 /// Logical filter operator that selects rows matching a condition.
 ///
 /// Takes input relation (`Relation`) and filters rows using a boolean predicate (`Scalar`).
 #[derive(Clone)]
 pub struct Filter<Relation, Scalar> {
+    /// The input relation.
     pub child: Relation,
+    /// The filter expression denoting the predicate condition for this filter operation.
```
yliang412, line 10: usually adding an example helps here than the verboseness.

e,g. `column_a > 10`

optd-core/src/operator/relational/logical/join.rs
```diff
@@ -1,11 +1,17 @@
...
 /// (`Scalar`).
 #[derive(Clone)]
 pub struct Join<Relation, Scalar> {
+    /// TODO(alexis) Mocked for now.
     pub join_type: String,
+    /// The left input relation.
     pub left: Relation,
+    /// The right input relation.
     pub right: Relation,
+    /// The join expression denoting the join condition that links the two input relations.
```
yliang412, line 15: Same here, giving an example helps.

yliang412 (approved):

AlSchlo (approved):

AlSchlo:
Perhaps we should rename logical scan to "get". Scan semantically sounds more like a physical implementation.

connortsui20:
> Perhaps we should rename logical scan to "get". Scan semantically sounds more like a physical implementation.

If we're going to rename it at all, it should probably be named `Read` instead of `Get`, but even so I think `Scan` is more universal and it will be clear what we're all talking about when we mention it.

PR #24, opened 2025-02-11
feat: DSL semantic checking analysis

AlSchlo:
## Problem
We want to reject DSL programs that are semantically incorrect. This PR analyzes the parsed AST and rejects if an error is detected.

## Summary of changes
- Parser bug fixes
- Implemented `SemanticAnalyzer` to perform comprehensive static analysis on DSL programs
- Added validation for properties:
  - Ensures property types are valid (basic types, maps, arrays, tuples)
  - Rejects function types and operators in properties
- Added validation for operators:
  - Prevents duplicate operator names
  - Validates field types for both scalar and logical operators
  - Ensures logical operators have all required derived properties
  - Verifies derived properties match declared logical properties
- Added scope-based identifier validation:
  - Maintains a stack of scopes for variable declarations
  - Prevents duplicate identifiers in the same scope
  - Validates variable references exist in accessible scopes
- Added recursive expression validation:
  - Validates all sub-expressions in compound expressions
  - Handles various expression types (variables, function calls, conditionals, etc.)
  - Manages scope for closures and pattern matching
- Added comprehensive test suite:
  - Tests for valid DSL programs
  - Tests for various error conditions (duplicate names, invalid types, etc.)
  - Tests for scope handling and identifier validation

The semantic analyzer provides early error detection for DSL programs, helping users identify issues before runtime.


optd-core/src/dsl/programs/example.optd
```diff
@@ -0,0 +1,148 @@
+// Shared properties for Logical operators
```
skyzh: dup w/ parser/programs?

optd-core/src/dsl/analyzer/semantic.rs
```diff
@@ -0,0 +1,575 @@
...
+                name: "field1".to_string(),
+                ty: Type::Function(Box::new(Type::Int64), Box::new(Type::Int64)),
+            }],
+        });
+
+        let mut analyzer = SemanticAnalyzer::new();
+        let result = analyzer.validate_file(&file);
+        assert!(result.is_err());
+        assert_eq!(
+            result.unwrap_err(),
```
skyzh, line 520: To make it user friendly in the future, we probably need to raise the compiler error at the site where the error occurs, for example, Line X Column Y, (something like a source code span could help, pest should support that)

skyzh (approved):
generally LGTM

optd-core/src/dsl/analyzer/semantic.rs
AlSchlo: Yes. I was planning to add that in the future. Usually you add that in the AST annotations.

PR #25, opened 2025-02-11
feat: DSL parser rewrite

AlSchlo:
## Problem
The previous parser was imperative, had poorly defined syntax, was missing the entire physical side, and was not maintainable (essentially a hackathon-like POC). This PR solves all of these problems by designing a better language and using a declarative parser.

## Summary of changes
- Better, more simplistic AST
- Support for ADTs
- Parsing with error recovery and useful messages
- Support map literals
- Much, much more expressive syntax to model query optimization (see test cases for now, will release a formal specification eventually)

## Next steps
- **Front-end:** Analyzer, type checker & type-polymorphism.
- **Back-end:** Write a HIR prototype and interpreter, linked to the Cascades engine. (<- will focus on this first)


optd-dsl/src/programs/working.optd
```diff
@@ -31,15 +31,6 @@ Logical Join(
   schema_len = left.schema_len + right.schema_len
 }
 
-Logical Sort(child: Logical, keys: [(Scalar, String)]) derive {
```
skyzh: consider support block comments in the language?

skyzh (approved):
looks like this would conflict with #26 and need to decide who merges first

optd-dsl/src/parser/adt.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    #[test]
+    fn test_double_nested_enum() {
+        let input = "data Menu with
+                         | File with
+                             | New with
+                                 | Document
+                                 | Project
+                                 \\ Template
```
yliang412, line 242: nit: Consider using raw string, i.e. `r#"<your test input>"# so we don't have to escape things.

yliang412 (approved):

PR #26, opened 2025-02-11
feat: full e2e pipeline

yliang412:
## Problem

With the initial representation and storage added in #4 and #22, we now want to support the full pipeline going from parsing SQL, optimizing the plan using optd, and executing the query in Datafusion.

## Summary of changes

- Integrate all @SarveshOO7's good work in https://github.com/cmu-db/optd/pull/10
- Added one mock physical implementation rule + operator for each logical operator
- Refactor scalar operator storage and reduce code bloat.
- Add physical storage tables and memo API.
- Bump MSRV to 1.81.0 to be compatible with datafusion 45.0.0: https://github.com/apache/datafusion/pull/14330


optd-core/src/cascades/memo.rs
```diff
@@ -67,4 +79,17 @@ pub trait Memoize: Send + Sync + 'static {
...
     ) -> Result<ScalarGroupId>;
+
+    /// Gets all logical expressions in a group.
+    async fn get_all_physical_exprs_in_group(
+        &self,
+        group_id: RelationalGroupId,
+    ) -> Result<Vec<(PhysicalExpressionId, Arc<PhysicalExpression>)>>;
+
+    /// Adds a physical expression to an existing group in the memo table.
+    async fn add_physical_expr_to_group(
```
skyzh, line 90: note that it's possible for a physical expr not to belong to any group with the optd fork design, so probably we need to think twice about the physexpr related apis

optd-core/src/cascades/mod.rs
```diff
@@ -64,7 +99,97 @@ pub async fn ingest_partial_scalar_plan(
...
+                mock_optimize_scalar_group(memo, *field).await?;
+            }
+        }
+    }
+
+    Ok(())
+}
+
+#[async_recursion]
+pub async fn match_any_partial_logical_plan(
```
skyzh, line 192: consider codegen this with the facilities in #25

optd-core/src/operators/relational/physical/project.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl<Relation, Scalar> PhysicalProject<Relation, Scalar> {
+    /// Create a new physical project operator.
+    pub fn new(child: Relation, fields: Vec<Scalar>) -> Self {
+        Self { child, fields }
+    }
+}
+
+/// Creates a physical project operator.
+pub fn project<Relation, Scalar>(
```
skyzh: naming: need to distinguish from logical project?

optd-core/src/operators/scalar/and.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use serde::Deserialize;
+
+use crate::values::OptdValue;
+
+use super::ScalarOperator;
+
+/// A scalar operator that adds two values.
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct And<Scalar> {
```
skyzh: I think we should collapse binary ops and logical ops into a single class like

```
pub struct LogicalOp<Scalar> {
  type: Or/And
}

pub struct BinaryOp<Scalar> {
  type: Add/Minus/...
}
```

which helps reduce the num of lines of code

optd-core/src/storage/memo.rs
```diff
@@ -518,6 +612,111 @@ impl SqliteMemo {
...
+        };
+        txn.commit().await?;
+
+        Ok(inserted_group_id)
+    }
+
+    /// Inserts an entry into the `physical_expressions` table.
+    async fn insert_into_physical_expressions(
+        txn: &mut SqliteConnection,
+        logical_expr_id: PhysicalExpressionId,
```
skyzh: rename to physical_expr_id

optd-core/src/storage/memo.rs
```diff
@@ -529,7 +728,21 @@ const fn get_all_logical_exprs_in_group_query() -> &'static str {
         " UNION ALL ",
         "SELECT logical_expression_id, json_object('Filter', json_object('child', child_group_id, 'predicate', predicate_group_id)) as data FROM filters WHERE group_id = $1",
         " UNION ALL ",
-        "SELECT logical_expression_id, json_object('Join', json_object('join_type', json(join_type), 'left', left_group_id, 'right', right_group_id, 'condition', condition_group_id)) as data FROM joins WHERE group_id = $1"
+        "SELECT logical_expression_id, json_object('Join', json_object('join_type', json(join_type), 'left', left_group_id, 'right', right_group_id, 'condition', condition_group_id)) as data FROM joins WHERE group_id = $1",
```
skyzh, line 732: another opportunity to codegen otherwise I can't imagine how many places we need to touch when adding an operator

optd-core/src/storage/memo.rs
```diff
@@ -544,7 +757,9 @@ const fn get_all_scalar_exprs_in_group_query() -> &'static str {
         " UNION ALL ",
         "SELECT scalar_expression_id, json_object('Add', json_object('left', left_group_id, 'right', right_group_id)) as data FROM scalar_adds WHERE group_id = $1",
         " UNION ALL ",
-        "SELECT scalar_expression_id, json_object('Equal', json_object('left', left_group_id, 'right', right_group_id)) as data FROM scalar_equals WHERE group_id = $1"
+        "SELECT scalar_expression_id, json_object('Equal', json_object('left', left_group_id, 'right', right_group_id)) as data FROM scalar_equals WHERE group_id = $1",
```
skyzh: I thought `And` will be represented as a list of expressions instead of left/right

optd-datafusion/src/converter/into_optd.rs
```diff
@@ -0,0 +1,176 @@
...
+                predicate: Self::conv_df_to_optd_scalar(
+                    &df_filter.predicate,
+                    df_filter.input.schema(),
+                    0,
+                )?,
+            }),
+            DFLogicalPlan::Join(join) => {
+                let mut join_cond = Vec::new();
+                for (left, right) in &join.on {
+                    let left = Self::conv_df_to_optd_scalar(left, join.left.schema(), 0)?;
```
skyzh, line 116: I would recommend make conversion + column ref rewrite into two passes because they are two different things

skyzh: Otherwise every scalar conversion function has to take an offset with it. Or you can make such offset into the context.

optd-datafusion/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+use datafusion::arrow::array::RecordBatch;
+use datafusion::arrow::util::pretty;
+use datafusion::physical_plan::ExecutionPlanProperties;
+use datafusion::physical_plan::Partitioning;
+use futures::StreamExt;
+use std::time::SystemTime;
+pub mod converter;
+pub mod planner;
+
+pub async fn run_queries(queries: String) -> Result<()> {
```
skyzh, line 26: Consider vendor datafusion-cli instead of implementing this by ourselves

optd-datafusion/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+        print_results.into_iter().for_each(|batch| {
+            let pretty_results = pretty::pretty_format_batches(&[batch]).unwrap().to_string();
+            println!("{}", pretty_results);
+        });
+    }
+    Ok(())
+}
+/// Utility function to create a session context for datafusion + optd.
+/// The allow deprecated is for the RuntimeConfig
+#[allow(deprecated)]
```
skyzh: We should use the new API from datafusion instead of marking allow deprecated

optd-datafusion/src/main.rs
```diff
@@ -0,0 +1,44 @@
+use datafusion::common::Result;
+use optd_datafusion::run_queries;
+use std::io;
+
+#[tokio::main]
```
skyzh: vendor datafusion-cli instead of implementing a less functional cli?

skyzh (approved):

optd-core/src/storage/memo.rs
yliang412: Yes, this needs to be codegened.

yliang412: created follow-up issue: https://github.com/cmu-db/optd/issues/27

optd-core/src/operators/relational/physical/project.rs
yliang412: 15fadcb77e043f16a879220da54e46232c3028c8

optd-core/src/operators/scalar/and.rs
yliang412: 05a856a8f72a2c86c51f74719ad8663e2c0c3da3

optd-core/src/storage/memo.rs
yliang412: 86a8fa6c9257e88d754f6759d292ed0707a0a1e8

optd-core/src/storage/memo.rs
yliang412: Redid the representation: 05a856a8f72a2c86c51f74719ad8663e2c0c3da3

optd-datafusion/src/lib.rs
yliang412: already use the newest API. I forgot to remove the flag. 91354733570993b6279a53effd1d2a3410c3bd8a

optd-datafusion/src/main.rs
yliang412: d533ede1875a6d0775a443f048fe1a857b272b4f

optd-datafusion/src/lib.rs
yliang412: d533ede1875a6d0775a443f048fe1a857b272b4f

yliang412: gonna remove this

optd-datafusion/src/converter/into_optd.rs
yliang412: Decide to do this in a later PR: https://github.com/cmu-db/optd/issues/28.

optd-core/src/cascades/memo.rs
yliang412: > it's possible for a physical expr not to belong to any group with the optd fork design

You mean the enforcer case?

I think I understand the problem. My solution would be to relax physical expression a little, saying that a physical expression does not have to be in the same group as the logical expression, it doesn't even need to be in any group, but it must be generated as a result of optimizing a logical expression + a required physical property. 

Yea then it doesn't make sense to have the `add_physical_expr_to_group` API.

yliang412: I will think about how to organize this into the memo table.

PR #34, opened 2025-02-22
feat: DSL Engine and HIR

AlSchlo:
# OPTD Engine Implementation

## Problem

This PR implements the OPTD engine that serves as the interface between the DSL and the optimizer driver. The engine provides the essential APIs for rule application, cost model integration, and catalog access that drive the exploration process. It enables efficient query plan transformations while managing the complexity of multiple possible transformation paths.

## Architecture

The engine exposes rule application to the optimizer driver:

```
┌───────────────────┐
│   Optimizer       │
│   Driver          │
└────────┬──────────┘
         │ match_and_apply_logical_rule("JoinCommutativity", plan)
         ▼
┌────────────────────────────────────────────────┐
│                  Engine                        │
├────────────────────────────────────────────────┤
│                                                │
│  ┌─────────────┐      ┌────────────────────┐   │
│  │ HIR Rules   │      │ Input Plan:        │   │
│  │ Repository  │      │ Join(A, B)         │   │
│  │             │      └────────────────────┘   │
│  │ function    │               │               │
│  │ JoinComm..  │               │               │
│  │ (plan) {...}│               ▼               │
│  └──────┬──────┘                               │
│         │         ┌────────────────────┐       │
│         └────────▶│ Evaluation Context │       │
│                   └──────────┬─────────┘       │
└───────────────────────────────│───────────────────┘
                                │
                                ▼
┌───────────────────────────────────────────────────┐
│           Expression Evaluation Stream             │
│                                                   │
│  ┌──────────────────┐    ┌──────────────────┐     │
│  │ Pattern Matching │    │ Rule Application │     │
│  │ Join(A, B)       │───▶│ Transformation   │     │
│  └──────────────────┘    └─────────┬────────┘     │
│                                   │               │
└───────────────────────────────────│───────────────┘
                                    │
                                    ▼
                          ┌─────────────────────┐
                          │ Transformed Plans   │
                          │                     │
                          │ ┌───────────────┐   │
                          │ │ Join(B, A)    │   │
                          │ └───────────────┘   │
                          │                     │
                          └─────────────────────┘
                                    │
                                    ▼
                          ┌─────────────────────┐
                          │ Back to Driver      │
                          │ (For cost-based     │
                          │  selection)         │
                          └─────────────────────┘
```

## Summary of Changes

- **Engine Interface**: Implemented the primary interface that the optimizer uses to apply rules to plans
- **HIR Implementation**: Core structures for expressions, patterns, operators, and values
- **Streaming Evaluation**: Non-blocking exploration of transformation paths
- **Bridge Module**: Bidirectional translation between HIR and Cascades IR

## Future Challenges

Performance optimization including: reducing copying footprint by implementing Arc-based references for expressions, contexts, and patterns; smart streams with caching capabilities to avoid redundant evaluations; and more efficient pattern matching for complex operator structures.


optd-core/src/cascades/ir.rs
```diff
@@ -0,0 +1,74 @@
...
+    Singleton(T),
+    VarLength(Vec<T>),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum PartialLogicalPlan {
+    PartialMaterialized {
+        tag: String,
+        data: Vec<OperatorData>,
+        relational_children: Vec<Children<PartialLogicalPlan>>,
```
skyzh, line 49: We can constrain the system in a way that we _either_ have fixed number of children _or_ have variable number of children to simplify the handling of children. Then we can have a `List` plan node type / predicate type in the system to be the only entity with variable number of children. (could be done in another patch)

In that way, we will have:

(LogicalProject (List expr1 expr2 expr3 ...) child)

(InExpr (List expr1 expr2 expr3 ...) child)

etc.

optd-core/src/cascades/properties/mod.rs
```diff
@@ -11,5 +11,5 @@ pub struct PhysicalProperties {
 pub struct SortProperty {
     /// Each tuple is a column index, direction pair.
     /// e.g. vec![(0, Asc), (1, Desc)]
-    pub sort_orders: Vec<(OptdValue, OptdValue)>,
+    pub sort_orders: Vec<(OperatorData, OperatorData)>,
```
skyzh, line 14: This was missed in the previous PR review: why using `OptdValue` instead of Vec<(usize, Order)> here? where usize = column index, Order is the enum of the sort. It reduces one layer of indirection.

optd-core/src/operators/relational/logical/join.rs
```diff
@@ -17,11 +17,11 @@ pub struct Join<Value, Relation, Scalar> {
     pub condition: Scalar,
 }
 
-impl<Relation, Scalar> Join<OptdValue, Relation, Scalar> {
+impl<Relation, Scalar> Join<OperatorData, Relation, Scalar> {
     /// Create a new join operator.
     pub fn new(join_type: &str, left: Relation, right: Relation, condition: Scalar) -> Self {
         Self {
-            join_type: OptdValue::String(join_type.into()),
+            join_type: OperatorData::String(join_type.into()),
```
skyzh, line 24: TODO: should be OperatorData::JoinType in the future

optd-core/src/operators/relational/logical/join.rs
```diff
@@ -17,11 +17,11 @@ pub struct Join<Value, Relation, Scalar> {
     pub condition: Scalar,
 }
 
-impl<Relation, Scalar> Join<OptdValue, Relation, Scalar> {
+impl<Relation, Scalar> Join<OperatorData, Relation, Scalar> {
```
skyzh, line 20: Why `Value` is part of the generic types? I suppose all plan nodes in the system will use the same operator data. If we split the crate into an optimizer framework and a user-defined set of plan nodes, both OperatorData and Join will be defined by the user, and users won't need to bother adding another layer of generics.

optd-core/src/operators/scalar/binary_op.rs
```diff
@@ -21,22 +21,26 @@ impl<Value, Scalar> BinaryOp<Value, Scalar> {
 }
 
 /// Creates an addition scalar operator.
-pub fn add<Scalar>(left: Scalar, right: Scalar) -> ScalarOperator<OptdValue, Scalar> {
-    ScalarOperator::BinaryOp(BinaryOp::new(OptdValue::String("add".into()), left, right))
+pub fn add<Scalar>(left: Scalar, right: Scalar) -> ScalarOperator<OperatorData, Scalar> {
```
skyzh, line 24: style: I'd prefer generic types to have single letter names, i.e., `add<S>(left: S, right: S)` so that it fits more into the Rust convention. We can resolve this later.

optd-dsl/Cargo.toml
```diff
@@ -7,4 +7,7 @@ edition = "2021"
 optd-core = { path = "../optd-core" }
 chumsky = "0.9.3"
 ariadne = "0.5.0"
-ordered-float = "4.6.0"
+ordered-float = "5.0.0"
+futures = "0.3.31"
+tokio.workspace = true
```
skyzh, line 12: style: consider making all dependencies workspace-level

optd-dsl/src/engine/eval/core.rs
```diff
@@ -0,0 +1,390 @@
...
+            stream_from_result(
+                keys_result,
+                capture!([values, context], move |keys| {
+                    // Then evaluate all value expressions
+                    evaluate_all_combinations(values.into_iter(), context)
+                        .map(capture!([keys], move |values_result| {
+                            // Create map from keys and values
+                            values_result.map(|values| {
+                                Value(CoreData::Map(
+                                    keys.clone().into_iter().zip(values).collect(),
```
skyzh: nit: might be more efficient to do `into_iter().cloned()`

optd-dsl/src/engine/eval/expr.rs
skyzh, line 1: Note: I didn't look deeply into the eval module as I believe bugs can easily be found by looking at the evaluation results of the programs so I'm not doing reviews on the correctness of the code.

optd-dsl/src/engine/bridge/from_optd.rs
skyzh, line 1: I wish there could be a single representation for both optd operators and their representation in the interpreter to avoid codegen too many things, i.e., it would be ideal if we can make partial logical plan a variant of the interpreter value.

optd-dsl/src/engine/utils/macros.rs
```diff
@@ -0,0 +1,42 @@
...
+/// // With the capture! macro:
+/// some_function(capture!([tag, context], move |x| {
+///     // The closure receives already-cloned variables
+///     process(tag, context, x);
+/// }));
+/// ```
+///
+/// This is particularly useful in async code with many nested closures
+/// that need to capture the same variables.
+#[macro_export]
```
skyzh, line 34: Alternative: https://crates.io/crates/closure; this macro might introduce too many copies. Also we have async closure in Rust edition 2024 and not sure if it would help.

optd-dsl/src/engine/utils/streams.rs
```diff
@@ -0,0 +1,189 @@
...
+/// # Type Parameters
+/// * I - The type of the iterator over expressions
+///
+/// # Parameters
+/// * items - Iterator over expressions to evaluate
+/// * context - Evaluation context
+///
+/// # Returns
+/// A stream of all possible combinations of values from the expressions
+pub(crate) fn evaluate_all_combinations<I>(mut items: I, context: Context) -> VecValueStream
```
skyzh, line 77: on the language capabilities: do we / can we support something like "if flat_map returns no value do something"? Seems like the function will return empty vector in such cases and therefore eliminating all states.

skyzh (approved):
Overall LGTM. This patch captures the core idea that rule matching DSL == executing a non-determinstic program and implemented necessary components to make this happen.

I'd like to see more e2e tests (rules) get implemented and see how it works with the cascades engine.

It seems like `expand_group` is not implemented yet and I suppose there's still some work to do before things can run end-to-end.

AlSchlo:
Thanks for the review.

Yes, it's a bit hard to test the combinatorial part of the code right now unless I mock the expand group code. The idea is to have the engine hold a reference of the optimizer driver, where calls can be done to execute new tasks (e.g. expanding a group, launching another rule, etc.)

optd-core/src/cascades/ir.rs
AlSchlo: This is what the Children wrapper does. Except it's 1 or variable.

AlSchlo: The work/reward to support fixed length more than one is not really worth it right now, as it'd have to be represented in the DSL too.

optd-core/src/cascades/properties/mod.rs
AlSchlo: We cannot just 1-1 map the DSL type system to the Operator values unfortunately. Maybe I'm wrong though? I thought it was better practice to separate the two.

Now whether properties should be allowed any type of the DSL or only OperatorData, I don't know yet.

optd-core/src/operators/relational/logical/join.rs
AlSchlo: Yeah, all these operators are going to be rewritten soon anyways. I just changed the file for the rename.

optd-core/src/operators/relational/logical/join.rs
AlSchlo: For no good reason. This is all going to be cleaned up soon.

AlSchlo: Used to be because we pattern matched on the value. But this happens in the engine now, with a more generic type system.

optd-core/src/operators/scalar/binary_op.rs
AlSchlo: Agreed. Same, deprecated code mostly.

optd-dsl/Cargo.toml
AlSchlo: Oh, what's the difference?

optd-dsl/src/engine/bridge/from_optd.rs
AlSchlo: I disagree with this though. I spent lots of time experimenting with types, and all my attempts to unite the DSL types with the optimizer types introduced ugly dependencies.

I found that having a bridge was the cleanest approach.

AlSchlo: Example: in the DSL, your operators are expressions, but in the driver, they are always values (i.e. no if/then else expressions inside them).

We could make one super generic type, but that just adds confusion in my opinion. Because the optimizer crate doesn't really care about that extra abstraction.

optd-dsl/src/engine/utils/macros.rs
AlSchlo: This is a capture problem. This is a limitation of Rust (in C++ you can specify exactly what you want to copy with =). My macro is much more compact than anything I could find online.

I don't think it creates too many copies though?

optd-dsl/src/engine/utils/streams.rs
AlSchlo: Not sure I understand? Why what that be helpful?

What I do plan on supporting is specifying in the DSL how deep you want to explore a rule. For example, you could call a rule inside another rule, and mention you only want 1 alternative of the inner rule. This can be very, very handy. You pretty much combine heuristic optimization with cost-based optimization.

AlSchlo: If flatmap returns nothing, then there are just no ways to apply the rule.

PR #41, opened 2025-03-10
feat: Streaming Optimizer

SarveshOO7:
After many rewrites + discussions + considerations.

Notes: 
- The engine needs a rewrite to be able to support CPS and execution forks.
- Storage should have versioning to allow for persistence and resource management.


optd-core/src/optimizer/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///
+    /// Each job (like property derivation) gets a unique ID to correlate
+    /// results with pending messages that depend on it.
+    next_dep_id: i64,
+
+    //
+    // Exploration tracking
+    //
+    /// Track which groups we've started exploring with transformation rules
+    exploring_groups: HashSet<GroupId>,
```
SarveshOO7, line 169: We need to use concurrent hashsets and hashmaps in order to ensure there are no race conditions

optd-core/src/optimizer/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    /// Track which goals we've started implementing with implementation rules
+    exploring_goals: HashSet<Goal>,
+
+    //
+    // Subscription management
+    //
+    /// Subscribers to logical expressions in groups
+    group_subscribers: HashMap<GroupId, Vec<Sender<LogicalExpression>>>,
```
SarveshOO7, line 178: concurrent hashmap needed

optd-core/src/optimizer/subscriptions.rs
```diff
@@ -0,0 +1,205 @@
...
+        sender: Sender<LogicalExpression>,
+    ) {
+        // Add the sender to group subscribers list
+        self.group_subscribers
+            .entry(group_id)
+            .or_default()
+            .push(sender.clone());
+
+        // Launch exploration if this group isn't being explored yet
+        if !self.exploring_groups.contains(&group_id) {
```
SarveshOO7, line 38: We have to make this atomic add

optd-core/src/optimizer/subscriptions.rs
```diff
@@ -0,0 +1,205 @@
...
+        sender: Sender<OptimizedExpression>,
+    ) {
+        // Add the sender to goal subscribers list
+        self.goal_subscribers
+            .entry(goal.clone())
+            .or_default()
+            .push(sender.clone());
+
+        // Launch exploration if this goal isn't being explored yet
+        if !self.exploring_goals.contains(&goal) {
```
SarveshOO7, line 79: have to make this atomic

optd-core/src/optimizer/mod.rs
AlSchlo: No, there is only one co-routine that accesses the state.

optd-core/src/optimizer/mod.rs
AlSchlo: Same

optd-core/src/optimizer/subscriptions.rs
AlSchlo: Likewise, not needed

optd-core/src/optimizer/subscriptions.rs
AlSchlo: Ditto

connortsui20 (approved):
Since this is getting super large, and since nobody else is working on this, let's just merge this now.

PR #42, opened 2025-03-13
feat: Engine CPS

AlSchlo:
Making the engine CPS so that it fits properly with the optimizer pub/sub model.

This is in fact a simplication of the codebase.


connortsui20 (approved):
I'm thinking we merge this immediately so that it doesn't block anyone else. It's not super useful for me to force the review loop now imo. Also, at a first glance there's nothing suspicious here anyways so I think this is fine to merge now.

What I'll probably do is make a separate PR that cleans up some of the documentation (mainly so I can get a better understanding of everything, but also to clean up).

yliang412 (approved):
Great to have the mock tests!

PR #43, opened 2025-03-14
Clean up doc comments and add periods

connortsui20:
## Problem

Cleans up a bit of the CPS PR.

## Summary of changes

This commit adds periods to all of the comments in the cir and engine. I also added some more detail here and there.

This commit also unifies all of the types in the cir to make it easier to interact with those types. Imo it makes more sense to have them all in one place rather than to have to figure out the paths to each of them every time you import.

### DIFF
--- a/optd-core/src/bridge/from_cir.rs
+++ b/optd-core/src/bridge/from_cir.rs
@@ -1,10 +1,6 @@
-use crate::cir::{
-    goal::Goal,
-    group::GroupId,
-    operators::{Child, OperatorData},
-    plans::{PartialLogicalPlan, PartialPhysicalPlan},
-    properties::{LogicalProperties, PhysicalProperties, PropertiesData},
-};
+//! Converts optd's type representations (CIR) into DSL [`Value`]s (HIR).
+
+use crate::cir::*;
 use optd_dsl::analyzer::hir::{
     self, CoreData, Literal, LogicalOp, Materializable, Operator, PhysicalOp, Value,
 };
@@ -13,19 +9,15 @@ use CoreData::*;
 use Literal::*;
 use Materializable::*;
 
-//=============================================================================
-// Main conversion functions
-//=============================================================================
-
-/// Converts a PartialLogicalPlan into a HIR Value representation.
+/// Converts a [`PartialLogicalPlan`] into a [`Value`].
 pub(crate) fn partial_logical_to_value(plan: &PartialLogicalPlan) -> Value {
     match plan {
         PartialLogicalPlan::UnMaterialized(group_id) => {
-            // For unmaterialized logical operators, we create a Value with the group ID
+            // For unmaterialized logical operators, we create a `Value` with the group ID.
             Value(Logical(LogicalOp(UnMaterialized(hir::GroupId(group_id.0)))))
         }
         PartialLogicalPlan::Materialized(node) => {
-            // For materialized logical operators, we create a Value with the operator data
+            // For materialized logical operators, we create a `Value` with the operator data.
             let operator = Operator {
                 tag: node.tag.clone(),
                 data: convert_operator_data_to_values(&node.data),
@@ -37,11 +29,11 @@ pub(crate) fn partial_logical_to_value(plan: &PartialLogicalPlan) -> Value {
     }
 }
 
-/// Converts a PartialPhysicalPlan into a HIR Value representation.
+/// Converts a [`PartialPhysicalPlan`] into a [`Value`].
 pub(crate) fn partial_physical_to_value(plan: &PartialPhysicalPlan) -> Value {
     match plan {
         PartialPhysicalPlan::UnMaterialized(goal) => {
-            // For unmaterialized physical operators, we create a Value with the goal
+            // For unmaterialized physical operators, we create a `Value` with the goal
             let hir_goal = cir_goal_to_hir(goal);
             Value(Physical(PhysicalOp(UnMaterialized(hir_goal))))
         }
@@ -58,7 +50,7 @@ pub(crate) fn partial_physical_to_value(plan: &PartialPhysicalPlan) -> Value {
     }
 }
 
-/// Converts LogicalProperties to a HIR Value representation.
+/// Converts [`LogicalProperties`] into a [`Value`].
 #[allow(dead_code)]
 pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Value {
     match &properties.0 {
@@ -67,7 +59,7 @@ pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Val
     }
 }
 
-/// Converts PhysicalProperties to a HIR Value representation.
+/// Converts [`PhysicalProperties`] into a [`Value`].
 pub(crate) fn physical_properties_to_value(properties: &PhysicalProperties) -> Value {
     match &properties.0 {
         Some(data) => properties_data_to_value(data),
@@ -75,11 +67,7 @@ pub(crate) fn physical_properties_to_value(properties: &PhysicalProperties) -> V
     }
 }
 
-//=============================================================================
-// CIR to HIR conversion helpers
-//=============================================================================
-
-/// Converts a CIR Goal to a HIR Goal.
+/// Converts a CIR [`Goal`] to a HIR [`Goal`](hir::Goal).
 fn cir_goal_to_hir(goal: &Goal) -> hir::Goal {
     let group_id = cir_group_id_to_hir(&goal.0);
     let properties = physical_properties_to_value(&goal.1);
@@ -90,12 +78,12 @@ fn cir_goal_to_hir(goal: &Goal) -> hir::Goal {
     }
 }
 
-/// Converts a CIR GroupId to a HIR GroupId.
+/// Converts a CIR [`GroupId`] to a HIR [`GroupId`](hir::GroupId).
 fn cir_group_id_to_hir(group_id: &GroupId) -> hir::GroupId {
     hir::GroupId(group_id.0)
 }
 
-/// Generic function to convert a Vec of Children to Vec of Values.
+/// A generic function to convert a slice of children into a vector of [`Value`]s.
 fn convert_children_to_values<T, F>(children: &[Child<Arc<T>>], converter: F) -> Vec<Value>
 where
     F: Fn(&T) -> Value,
@@ -112,12 +100,12 @@ where
         .collect()
 }
 
-/// Converts a slice of OperatorData to a Vec of HIR Values.
+/// Converts a slice of [`OperatorData`] into a vector of [`Value`]s.
 fn convert_operator_data_to_values(data: &[OperatorData]) -> Vec<Value> {
     data.iter().map(operator_data_to_value).collect()
 }
 
-/// Converts an OperatorData to a HIR Value representation.
+/// Converts an [`OperatorData`] into a [`Value`].
 fn operator_data_to_value(data: &OperatorData) -> Value {
     match data {
         OperatorData::Int64(i) => Value(Literal(Int64(*i))),
@@ -132,7 +120,7 @@ fn operator_data_to_value(data: &OperatorData) -> Value {
     }
 }
 
-/// Converts a PropertiesData to a HIR Value representation.
+/// Converts a [`PropertiesData`] into a [`Value`].
 fn properties_data_to_value(data: &PropertiesData) -> Value {
     match data {
         PropertiesData::Int64(i) => Value(Literal(Int64(*i))),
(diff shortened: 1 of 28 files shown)


optd-core/src/bridge/from_cir.rs
```diff
@@ -67,19 +59,15 @@ pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Val
     }
 }
 
-/// Converts PhysicalProperties to a HIR Value representation.
+/// Converts [`PhysicalProperties`] into a [`Value`].
```
SarveshOO7, line 62: let's make this value also a link to the actual type so that in the docs people can just click on it.

optd-core/src/cir/expressions.rs
```diff
@@ -6,76 +6,54 @@ use super::{
...
-/// representing a compact form suitable for the memo structure.
+/// Logical expressions have [`GroupId`]s as children rather than full plans, representing a compact
+/// form suitable for the memo structure.
 pub type LogicalExpression = Operator<GroupId>;
 
 /// A physical expression in the memo structure.
 ///
-/// Physical expressions use goal IDs rather than full plans for
-/// their children, representing a compact form suitable for the memo structure.
+/// Physical expressions use [`Goal`]s rather than full plans for their children, representing a
```
SarveshOO7, line 17: good catch

optd-core/src/cir/expressions.rs
```diff
@@ -6,76 +6,54 @@ use super::{
...
-    /// This creates a materialized partial plan where each child group ID
-    /// is converted to an unmaterialized partial plan reference.
+    /// This creates a materialized partial plan where each child group ID is converted to an
+    /// unmaterialized partial plan reference.
     fn from(expr: LogicalExpression) -> Self {
         PartialLogicalPlan::Materialized(Operator {
             tag: expr.tag,
             data: expr.data,
-            children: convert_children(expr.children),
+            children: convert_children::<GroupId, Self>(expr.children),
```
SarveshOO7, line 34: Nit: no need to specify types here

optd-core/src/cir/expressions.rs
```diff
@@ -6,76 +6,54 @@ use super::{
...
-/// This handles both singleton and variable-length children, converting
-/// each source type into its equivalent target wrapped in an Arc.
-///
-/// # Type Parameters
-/// * `S` - Source type (e.g., GroupId)
-/// * `T` - Target type (e.g., PartialLogicalPlan)
-///
-/// # Arguments
-/// * `children` - Collection of child items to convert
+/// A generic function that converts children with type `S` into a `Arc<T>`.
```
SarveshOO7, line 53: I think I prefer the old comment here.

SarveshOO7: the eg. of what S and T can be is helpful

optd-core/src/cir/goal.rs
```diff
@@ -1,11 +1,12 @@
 use super::{group::GroupId, properties::PhysicalProperties};
 
-/// A physical optimization goal, consisting of a group to optimize and
-/// the required physical properties.
+/// A physical optimization goal, consisting of a group to optimize and the required physical
```
SarveshOO7, line 3: these changes are unnecessary?

optd-core/src/cir/group.rs
```diff
@@ -1,7 +1,12 @@
-/// Unique identifier for a group in the memo structure.
+/// A unique identifier for a group in the memo structure.
```
SarveshOO7, line 1: Unnecessary

optd-core/src/cir/operators.rs
```diff
@@ -1,51 +1,52 @@
 use ordered_float::OrderedFloat;
```
SarveshOO7, line 2: this is also mostly feels unnecessary.

optd-core/src/cir/plans.rs
```diff
@@ -5,32 +5,33 @@ use super::{
...
 pub struct PhysicalPlan(pub Operator<Arc<PhysicalPlan>>);
 
-//=============================================================================
-// Partial Plan Types
-//=============================================================================
-
-/// Represents a partially materialized logical plan.
+/// Represents partially materialized logical plans.
+///
+/// Partial logical plans are partially materialized [`LogicalPlan`]s, representing the top of a
```
SarveshOO7: The Partial logical plan datatype does not represent the top of a logical plan tree. This is wrong/misleading/hard to understand depending on what you actually meant to say.

optd-core/src/cir/plans.rs
```diff
@@ -5,32 +5,33 @@ use super::{
...
+/// logical plan tree, where the unmaterialized leaves of this structure are [`GroupId`]s.
 ///
-/// Partial logical plans can either be fully materialized with operator trees,
-/// or unmaterialized references to logical groups in the memo structure.
+/// We use partial plans because most rules only need to match and bind to the top of the
+/// plan (the root and maybe some of its children), and it would be expensive to materialize the
+/// entire plan when the rule engine only needs to know about the top. In other words, partial plans
+/// allow us to give the rule engine exactly what it needs and nothing else.
```
SarveshOO7, line 42: Other than the last line of this paragraph, I think most of this should be changed. We can take this offline.

optd-core/src/cir/plans.rs
```diff
@@ -41,20 +42,21 @@ pub enum PartialLogicalPlan {
 
 /// Represents a partially materialized physical plan.
 ///
-/// Partial physical plans can either be fully materialized with operator trees,
-/// or unmaterialized references to physical goals.
+/// Partial physical plans are partially materialized [`PhysicalPlan`]s, representing the top of a
+/// physical execution plan tree, where the unmaterialized leaves of this structure are [`Goal`]s.
```
SarveshOO7: i think you should say "the top of a partial physical plan" tree

optd-core/src/cir/plans.rs
```diff
@@ -90,29 +92,15 @@ impl From<PhysicalPlan> for PartialPhysicalPlan {
...
-/// This handles both singleton and variable-length children, recursively
-/// converting each child plan to its partial equivalent.
-///
-/// # Type Parameters
-/// * `S` - Source plan type
-/// * `T` - Target partial plan type
-///
-/// # Arguments
-/// * `children` - Collection of child plans to convert
+/// A generic function that converts children with type `Arc<S>` into a `Arc<T>`.
```
SarveshOO7, line 116: I prefer the old documentation here as well because it has more detail

optd-core/src/cir/properties.rs
```diff
@@ -4,36 +4,36 @@ use super::plans::LogicalPlan;
...
 ///
-/// Physical properties describe execution characteristics of a physical plan node,
-/// such as sort order, distribution, or resource requirements.
+/// Physical properties describe physical execution characteristics of a physical plan node, such as
+/// sort order, distribution, or resource requirements.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct PhysicalProperties(pub Option<PropertiesData>);
 
-/// Represents various types of property data that can be associated with plan nodes.
+/// Represents various types of property data that can be associated with execution operators.
```
SarveshOO7: there is no concept of execution operators. If you meant to say physical operators then introducing redundant terminology is harmful to overall readability. Furthermore, PropertiesData can be associated with LogicalProperties as well.

optd-core/src/engine/eval/binary.rs
```diff
@@ -1,40 +1,32 @@
 //! This module provides implementation for binary operations between values.
 //!
-//! Binary operations are fundamental for expression evaluation, allowing values
-//! to be combined and compared in various ways. The module supports operations on
-//! different value types, including:
-//! - Arithmetic operations on numbers (add, subtract, multiply, divide)
-//! - Comparison operations for various types (equality, less than)
-//! - Logical operations on boolean values (AND, OR)
-//! - Collection operations (concatenation, range creation)
+//! Binary operations are fundamental for expression evaluation, allowing values to be combined and
```
SarveshOO7, line 3: unnecessary

optd-core/src/engine/eval/binary.rs
```diff
@@ -1,40 +1,32 @@
...
+//! - Logical operations on boolean values (AND, OR).
+//! - Collection operations (concatenation, range creation).
 
 use optd_dsl::analyzer::hir::{BinOp, CoreData, Literal, Value};
 
 /// Evaluates a binary operation between two values.
 ///
-/// This function performs binary operations on values based on their types
-/// and the operation requested. It handles arithmetic, logical, comparison,
-/// and collection operations.
```
SarveshOO7, line 17: why get rid of this documentation

optd-core/src/engine/eval/binary.rs
```diff
@@ -45,12 +37,12 @@ pub(super) fn eval_binary_op(left: Value, op: &BinOp, right: Value) -> Value {
                 _ => unreachable!(), // This branch is unreachable due to pattern guard
             })),
 
-            // Integer range operation (creates an array of sequential integers)
+            // Integer range operation (creates an array of sequential integers).
             (Int64(l), Range, Int64(r)) => {
-                Value(Array((*l..=*r).map(|n| Value(Literal(Int64(n)))).collect()))
+                Value(Array((l..=r).map(|n| Value(Literal(Int64(n)))).collect()))
```
SarveshOO7, line 42: Nit: I believe the * helps with readability but I know this is an unpopular opinion so do with that what you wish.

SarveshOO7 (changes requested):

optd-core/src/bridge/from_cir.rs
connortsui20: It is a link to the actual type, do you mean a link to the file while you are looking at the source code?

optd-core/src/cir/expressions.rs
connortsui20: No this was a nit on my end, if you don't specify the type it is very difficult to understand what is actually happening here. It took me about 3 minutes to figure this out even with an LSP

optd-core/src/cir/expressions.rs
connortsui20: In general I dislike comments that explain the function signature essentially verbatim, they don't really add anything in helping someone understand _why_ this function is here and ends up cluttering headspace.

optd-core/src/cir/plans.rs
connortsui20: If that's not the case, then I think I have fundamental misunderstanding of what a partial plan is. How is it not the top of a plan tree?

optd-core/src/engine/eval/binary.rs
connortsui20: I don't really see how adding operators that don't need to be there helps with readability... Make sure you look up top to see that I removed the references that didn't need to be there

optd-core/src/cir/goal.rs
connortsui20: It's just aligning to the 100 column limit standard

optd-core/src/engine/eval/binary.rs
connortsui20: See above, explaining simple function signatures is unnecessary

optd-core/src/cir/group.rs
connortsui20: I only see this as a net positive

optd-core/src/cir/operators.rs
connortsui20: Code order is actually a pretty important thing, you have to consider that people are reading your code for the first time, and thus you should probably put the most important things at the top of the file.

optd-core/src/cir/plans.rs
connortsui20: whoops you are right

optd-core/src/cir/plans.rs
connortsui20: See above

optd-core/src/engine/eval/binary.rs
connortsui20: Again, this is just the standard. Or are you saying that the module documentation is unnecessary?

SarveshOO7: I meant changing the comments to be 100 characters per line instead of what I presume what it was earlier: 80 characters

optd-core/src/cir/operators.rs
SarveshOO7: I meant moving words around just because we want to have 100 character line limits instead of 80 or whatever it was earlier

SarveshOO7 (approved):

optd-core/src/cir/operators.rs
connortsui20: https://doc.rust-lang.org/nightly/style-guide/#indentation-and-line-width

AlSchlo (approved):

PR #45, opened 2025-03-26
Unify Continuations

connortsui20:
## Problem

There were a lot of type aliases that looked identical, so I removed them for this:

```rust
pub type Continuation<Input> = Arc<dyn Fn(Input) -> UnitFuture + Send + Sync + 'static>;
```

## Summary of changes

Removed all of the `???Continuation` type aliases

I also lifted the `Generator` trait into the same module as `engine`

### DIFF
--- a/optd-core/src/engine/mod.rs
+++ b/optd-core/src/engine/mod.rs
@@ -11,8 +11,7 @@ use crate::{
     capture,
     cir::{Cost, LogicalProperties, PartialLogicalPlan, PartialPhysicalPlan, PhysicalProperties},
 };
-use eval::Evaluate;
-use generator::Generator;
+
 use optd_dsl::analyzer::{
     context::Context,
     hir::{CoreData, Expr, Literal, Value},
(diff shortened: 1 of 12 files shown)


optd-core/src/engine/mod.rs
```diff
@@ -22,25 +21,22 @@ use utils::UnitFuture;
...
-pub(crate) type CostContinuation = Arc<dyn Fn(Cost) -> UnitFuture + Send + Sync + 'static>;
+#[cfg(test)]
+pub(super) mod test_utils;
 
-/// Type alias for a continuation that receives LogicalProperties
-pub(crate) type PropertiesContinuation =
-    Arc<dyn Fn(LogicalProperties) -> UnitFuture + Send + Sync + 'static>;
+/// A type alias for continuations used in the rule engine.
+///
+/// The engine uses continuation-passing-style (CPS) since it requires advanced control flow to
```
AlSchlo, line 36: Slightly confusing wording, but there are many other places we have a bit of confusing documentation.
I'll do a full codebase audit and fix these nits in 2 months.

AlSchlo (approved):
Thanks for fixing this. I struggled with it and thought it would need nightly. Good job.

PR #47, opened 2025-03-27
Refactor rule engine + migrate to `optd-dsl`

connortsui20:
## Problem

We want to have the `engine` module located inside the `optd-dsl` crate so that we don't end up with a circular dependency when trying to allow the dsl "UDF"s to access the `Engine`.

```rust
// Right now (defined in `optd-dsl`):
RustUDF(fn(Vec<Value>) -> Value),

// We want something similar to:
RustUDF(fn(Engine, Vec<Value>) -> Value)
//         ^^^^^^ If this was in the `optd-core` crate we would have a circular dependency
```

## Summary of changes

There was a lot of duplicate code and unnecessary function signatures, so I cleaned all of it up.

Additionally, the entire `engine` module was moved into `optd-dsl`. This caused a lot of problems, so a lot of function signatures had to be changed.

Notably, I got rid of the `Evaluate` trait inside `optd-core` and added an `evaluate` method directly on `Engine`. I still think that is would be good to have every `evaluate_???` function be a private method on `Engine` for readability purposes, but that can be a different PR.

(It might be good to look at the individual commits rather than the entire diff)


optd-dsl/src/engine/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    ///
+    /// * `self` - The evaluation engine (owned)
+    /// * `expr` - The expression to evaluate
+    /// * `k` - The continuation to receive each evaluation result
+    pub fn evaluate(
+        self,
+        expr: Arc<Expr>,
+        k: Continuation<Value>,
+    ) -> impl Future<Output = ()> + Send {
+        use eval::core::evaluate_core_expr;
```
AlSchlo: Why are these inside the function?

connortsui20: Whoops, forgot to move that out

AlSchlo (changes requested):
Good job. However I don't like the the imports inside the functions. That really clutters the function code.

connortsui20:
I also moved `capture` out to `lib.rs` and the two future related types into the `engine/utils.rs`. I think the names aren't the best but it's not that important right now

optd-core/src/optimizer/handlers.rs
```diff
@@ -169,9 +169,16 @@ impl<M: Memoize> Optimizer<M> {
...
 
-                // Launch the cost plan operation with the continuation
+                // Launch the cost plan operation with the continuation.
                 tokio::spawn(async move {
-                    engine.launch_cost_plan(&plan, continuation).await;
+                    engine
+                        .launch_rule(
+                            "cost",
+                            vec![partial_physical_to_value(&plan)],
+                            crate::bridge::into_cir::value_to_cost,
```
AlSchlo: Please import

optd-core/src/optimizer/subscriptions.rs
```diff
@@ -198,10 +210,15 @@ impl<M: Memoize> Optimizer<M> {
...
 
-                    // Launch the cost plan operation with the continuation
+                    // Launch the cost plan operation with the continuation.
                     engine_clone
-                        .launch_cost_plan(&plan, cost_continuation)
-                        .await;
+                        .launch_rule(
+                            "cost",
+                            vec![partial_physical_to_value(&plan)],
+                            crate::bridge::into_cir::value_to_cost,
```
AlSchlo: Ditto

yliang412 (approved):
LGTM

optd-core/src/optimizer/subscriptions.rs
AlSchlo: Fixed

optd-core/src/optimizer/handlers.rs
AlSchlo: Fixed

AlSchlo (approved):

PR #48, opened 2025-03-29
feat: Generic CPS

yliang412:
## Problem

We need a way to finish a job on a tokio task and use the correct job ID to send a message back to the main event loop. Before this PR, we always returned the unit type for continuations. We achieve correctness while maintaining good abstractions between the engine and the optimizer by making continuation return an `EngineResponse<O>` with a converter function into some output type.

```rust
/// The engine response type, which can be either a return value with a converter callback
/// or a yield group/goal with a continuation for further processing.
pub enum EngineResponse<O> {
    /// The engine has returned a value, and the final continuation
    /// should be called on it to produce the desired output.
    Return(Value, Continuation<Value, O>),
    /// The engine has yielded a group ID, and the continuation
    /// should be called for each value after expanding the group.
    YieldGroup(GroupId, Continuation<Value, EngineResponse<O>>),
    /// The engine has yielded a goal, and the continuation
    /// should be called for each value after expanding the goal.
    YieldGoal(Goal, Continuation<Value, EngineResponse<O>>),
}
```

## Summary of changes

- Remove the generator, we don't need it anymore.
- No longer needs to pass channels to the engine.
  - We can construct the `OptimizerMessage` using the final callback, but send messages outside the chain of continuations.
- Keeping the initial callback (or we call it the `transform` continuation):
  -  We use it for validating the output `Value` to CIR representation and constructing the single message type.
- The output type is made generic so `optd-dsl` does not have recursive dependencies.

### Notes for reviewers

New signatures for the future and continuation:

https://github.com/cmu-db/optd/blob/272fc6842213c7766f9824f5dbd298fa8b19ff2b/optd-dsl/src/engine/utils.rs#L10-L17

The only place we return a `YieldGroup`:

https://github.com/cmu-db/optd/blob/272fc6842213c7766f9824f5dbd298fa8b19ff2b/optd-dsl/src/engine/eval/match.rs#L182-L189

The only place we return a `YieldGoal`:

https://github.com/cmu-db/optd/blob/272fc6842213c7766f9824f5dbd298fa8b19ff2b/optd-dsl/src/engine/eval/match.rs#L193-L199

**Note:** The state of the optimizer on `main` is very behind what we have right now so I commented a lot of stuff out, but I made all engine tests pass and thought about the cases for integrating it with the optimizer.


optd-core/src/optimizer/handlers.rs
```diff
@@ -149,37 +144,36 @@ impl<M: Memoize> Optimizer<M> {
             }
 
             // If a new expression was created, cost it using CPS
-            if let Some(expr) = new_expr {
-                let message_tx = self.message_tx.clone();
-                let engine = self.engine.clone();
+            if let Some(_expr) = new_expr {
+                // let message_tx = self.message_tx.clone();
+                // let engine = self.engine.clone();
```
connortsui20: Could you just make this a `todo!()`?

optd-core/src/optimizer/ingest.rs
```diff
@@ -60,39 +54,38 @@ impl<M: Memoize> Optimizer<M> {
...
+                    .map(|_expr| {
                         let job_id = self.next_dep_id;
                         self.next_dep_id += 1;
 
-                        let message_tx = self.message_tx.clone();
-                        let engine = self.engine.clone();
-                        let expr_clone = expr.clone();
+                        // let message_tx = self.message_tx.clone();
+                        // let engine = self.engine.clone();
+                        // let expr_clone = expr.clone();
```
connortsui20: Same here, just make these `todo!()`s, we can always get this code back from git

optd-dsl/src/engine/utils.rs
```diff
@@ -1,30 +1,34 @@
...
-pub type UnitFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
+/// A type alias for a future that completes with return value of type `T`.
+pub type PinnedFuture<T> = BoxFuture<'static, T>;
 
 /// A type alias for continuations used in the rule engine.
 ///
 /// The engine uses continuation-passing-style (CPS) since it requires advanced control flow to
 /// expand (enumerate) expressions within a group.
-pub type Continuation<Input> = Arc<dyn Fn(Input) -> UnitFuture + Send + Sync + 'static>;
+pub type Continuation<I, O> = Arc<dyn Fn(I) -> PinnedFuture<O> + Send + Sync + 'static>;
```
connortsui20: You can remove the `'static` bound here if the return value `PinnedFuture` is `'static` (or if it becomes `BoxFuture<'static, O>`.

optd-dsl/src/engine/mod.rs
```diff
@@ -100,26 +109,23 @@ impl<G: Generator> Engine<G> {
...
         name: &str,
         values: Vec<Value>,
-        transform: fn(&Value) -> T,
-        k: Continuation<T>,
-    ) where
-        T: 'static,
+        return_k: Continuation<Value, O>,
+    ) -> EngineResponse<O>
+    where
+        O: Send + 'static,
```
connortsui20, line 129: Just a question: if I understand this correctly, this basically means that it is on the optimizer to call the continuations once it is done processing the return from the rule engine?

optd-dsl/src/engine/utils.rs
connortsui20: Given that there is clearly a much tighter coupling between the optimizer and the rule engine now, I think it would be good to consider:
- Moving these type aliases up to the crate level `lib.rs`
- Renaming the `Engine` to `RuleEngine`

optd-dsl/src/engine/utils.rs
```diff
@@ -1,30 +1,34 @@
...
-use std::{future::Future, pin::Pin, sync::Arc};
+use crate::engine::Engine;
+use std::sync::Arc;
+
+use super::EngineResponse;
 
-/// A type alias for a future that completes with no return value.
-pub type UnitFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
+/// A type alias for a future that completes with return value of type `T`.
+pub type PinnedFuture<T> = BoxFuture<'static, T>;
```
connortsui20: I think I would prefer to not see this type aliased. Either that or a come up with a different name for this? I feel that a type alias here makes things more confusing since it it hiding the `'static` away (which is important to understand)

optd-dsl/src/engine/eval/match.rs
```diff
@@ -120,156 +119,135 @@ where
...
-    k: Continuation<MatchResult>,
-) -> UnitFuture
+    k: Continuation<MatchResult, EngineResponse<O>>,
+) -> PinnedFuture<EngineResponse<O>>
 where
-    G: Generator,
+    O: Send + 'static,
 {
     Box::pin(async move {
         match (pattern, &value.0) {
```
connortsui20, line 134: Missed this, this doesn't need to be a reference

connortsui20 (changes requested):
Looks good! I think this model makes a lot more sense and is more general, but we should be aware that now the engine is much more tightly coupled to the optimizer (because it relies on the optimizer to fire the continuations for the engine).

I think that the struct itself should probably be renamed to `RuleEngine` for clarity, but that's just my personal opinion.

Also, I think just add a `todo!()` on top of everything you commented out for now

optd-dsl/src/engine/mod.rs
yliang412: Yes.  I added this because we want to convert the results to different types (partial logical, partial physical, cost) once the engine finishes execution. 

If we don't do this, we need to manually check `Value` returned by the engine, and based on the enum type construct messages. In my opinion this is more error-prone, since we lose the initial intent.

An alternative I considered is that we can define a unified type that the optimizer expects, but it feels you have circular dependency between the two crates.

optd-dsl/src/engine/utils.rs
yliang412: In a way, the continuation does not have to come from the optimizer, right? The test case I modified shows how you could use the engine without the optimizer. 

Renaming it to rule engine might be helpful.

optd-dsl/src/engine/eval/match.rs
yliang412: There are a lot of places that take a reference to value in the matching code, and then `clone` is the callee. Feels like a lot of them should pass by value?

connortsui20: Yeah I think a lot of the function signatures need to change in general, but I was going to leave that to another PR

optd-dsl/src/engine/utils.rs
yliang412: removed the alias and make 'static bound explicit using `BoxFuture`: b04ba67cf5bfb9241b5bf10b4d515765ecdfd926

optd-dsl/src/engine/utils.rs
yliang412: b04ba67cf5bfb9241b5bf10b4d515765ecdfd926

optd-dsl/src/engine/eval/match.rs
yliang412: I will just leave it to another PR.

optd-core/src/optimizer/handlers.rs
yliang412: replaced stale code with todos everywhere: c9f68425c4501a2e6f0be9aee8f460dce4925089

optd-core/src/optimizer/ingest.rs
yliang412: c9f68425c4501a2e6f0be9aee8f460dce4925089

optd-dsl/src/engine/utils.rs
yliang412: Gonna leave it out. Since the engine technically will also be able to derive properties and compute cost. I would do the rename if it is only applying transformation and implementation rules.

connortsui20 (approved):

optd-dsl/src/engine/mod.rs
AlSchlo: Yes. We run into the circular dependency problem. I  believe this approach is probably the best. Because having to call the bridge from the value is as you said it pretty error prone (control flow is scattered).

optd-dsl/src/engine/utils.rs
AlSchlo: Yes. Engine should remain being called like that. It can do other stuff.

AlSchlo (dismissed):
I believe we should not contain a continuation in the return type.
As Connor said, this requires that the optimizer calls the continuation, which is not ideal.

We can take advantage from the fact we only build a Return type in `launch_rule`. Just apply the `transform` in there. Let's also not call it `continuation`.

Thoughts? @connortsui20 @yliang412 

```rust
self.evaluate(
            rule_call,
            Arc::new(move |result| {
                Box::pin(async move { EngineResponse::Return(transform(result)) })
            }),
        )
```

connortsui20:
> I believe we should not contain a continuation in the return type. As Connor said, this requires that the optimizer calls the continuation, which is not ideal.
> 
> We can take advantage from the fact we only build a Return type in `launch_rule`. Just apply the `transform` in there. Let's also not call it `continuation`.
> 
> Thoughts? @connortsui20 @yliang412
> 
> ```rust
> self.evaluate(
>             rule_call,
>             Arc::new(move |result| {
>                 Box::pin(async move { EngineResponse::Return(transform(result)) })
>             }),
>         )
> ```

Do you mean you don't like the entire `EngineResponse` struct? There are other places where the optimizer would need to call the continuation with this new `EngineResponse` return type.

I understand that you want the engine here to be completely abstracted from the optimizer, but I think that might be a bit of wishful thinking: I think the engine needs to have a very easy way of communicating with the optimizer. Really, I think it makes sense that the optimizer should be in charge of deciding which continuations get fired and which don't. It has to do this anyways for groups and goals, why not make it so it can do that with anything?

This also leads into my thoughts about the scheduler / async executor (which TLDR I believe quite strongly that a custom executor is not only necessary but should be a priority, I can summarize my discussion with @yliang412 if you want later). In the future, I think it would be a huge problem if the optimizer had limited control over what continuations were getting fired. I think it's fine that the engine can build up the continuations and then pass the control flow back to the optimizer so it can decide if it's actually going to fire them. Remember that the whole point of CPS here was that it would be hard for the optimizer itself to deal with the non-local control flow of expanding groups - I think the goal should be for the engine to support that in as minimal a way as possible, rather than try to do everything.

connortsui20:
Just got off a call with alexis, I think this change makes sense for now since this is the only thing we want that is different from groups or goals that should return control back to the optimizer. I still think that in the future there should be a way (other than by yielding a group or goal) for the engine to return control back, but that is a story for another time

AlSchlo:
After call with Yuchen: having continuations in the Return type is OK.
Problem is that we never really know what will come out of an engine result. Could be forked or not.

So this means that we would have to analyze what it is every time. Annoying...

Putting everything inside a closure is less error prone. I guess we should add more documentation about that later on.
It's very understandable why we do it this way, but easy to forget and get confused.

PR #49, opened 2025-03-31
event loop optimizer initial setup

yliang412:
I'm trying to get Alexis's existing WIP into main, so we don't have a merging nightmare.

Many constructs changed from the main over the past two weeks, and we want to use them as well in the newest rewrite.


AlSchlo (approved):
Most of this work is still WIP.
But I'm fine with merging this already, as the old code was very much stale.

PR #50, opened 2025-04-01
(feat) DSL: add optionals & logical/physical subtypes

AlSchlo:
This PR does two things:

1. Add optional types in the DSL to be able to return `None` when the rule application fails for example. 

Adapted the parser accordingly, as well as the HIR. See syntax below for types now.

```
/// Syntax examples:
/// - I64                   - Integer type
/// - String                - String type
/// - [I64]                 - Array of integers
/// - {String : I64}        - Map from strings to integers
/// - (I64, String)         - Tuple with integer and string
/// - I64 -> String         - Function from integer to string
/// - I64?                  - Optional integer
/// - [String]?             - Optional array of strings
/// - I64 -> String?        - Function returning optional string
/// - (I64 -> String)?      - Optional function
```

2. We need a type hierarchy in the DSL for Logical & Physical operators. At some places, we want to be able to access 

Specifically, we define:

```
StoredLogical: Operator + GroupId
Logical: Operator

StoredLogical <: Logical
```

and,

```
CostedPhysical: Operator + Goal (i.e. GroupId + Props) + Cost (i.e. f64 + stats in the future)
StoredPhysical: Operator + Goal
Physical: Operator

CostedPhysical <: StoredPhysical <: Physical
```

This translates as follows in the HIR.

```
pub struct LogicalOp<T> {
    pub operator: Materializable<Operator<T>, GroupId>,
    pub group_id: Option<GroupId>,
}
pub struct PhysicalOp<T> {
    pub operator: Materializable<Operator<T>, Goal>,
    pub goal: Option<Goal>,
    pub cost: Option<Box<Value>>,
}
```

### DIFF
--- a/optd-core/src/bridge/from_cir.rs
+++ b/optd-core/src/bridge/from_cir.rs
@@ -65,7 +65,7 @@ pub(crate) fn costed_physical_to_value(plan: PartialPhysicalPlan, cost: Cost) ->
 pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Value {
     match &properties.0 {
         Some(data) => properties_data_to_value(data),
-        None => Value(Null),
+        Option::None => Value(None),
     }
 }
 
@@ -73,7 +73,7 @@ pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Val
 pub(crate) fn physical_properties_to_value(properties: &PhysicalProperties) -> Value {
     match &properties.0 {
         Some(data) => properties_data_to_value(data),
-        None => Value(Null),
+        Option::None => Value(None),
     }
 }
 
--- a/optd-core/src/bridge/from_cir.rs
+++ b/optd-core/src/bridge/from_cir.rs
@@ -14,7 +14,9 @@ pub(crate) fn partial_logical_to_value(plan: &PartialLogicalPlan) -> Value {
     match plan {
         PartialLogicalPlan::UnMaterialized(group_id) => {
             // For unmaterialized logical operators, we create a `Value` with the group ID.
-            Value(Logical(LogicalOp(UnMaterialized(hir::GroupId(group_id.0)))))
+            Value(Logical(LogicalOp::logical(UnMaterialized(hir::GroupId(
+                group_id.0,
+            )))))
         }
         PartialLogicalPlan::Materialized(node) => {
             // For materialized logical operators, we create a `Value` with the operator data.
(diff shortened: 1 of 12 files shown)


optd-core/src/bridge/from_cir.rs
```diff
@@ -65,15 +67,15 @@ pub(crate) fn costed_physical_to_value(plan: PartialPhysicalPlan, cost: Cost) ->
 pub(crate) fn logical_properties_to_value(properties: &LogicalProperties) -> Value {
     match &properties.0 {
         Some(data) => properties_data_to_value(data),
-        None => Value(Null),
+        Option::None => Value(None),
```
yliang412, line 68: nit: I don't like that `CoreData::None` is polluting the namespace.

optd-dsl/src/engine/eval/match.rs
```diff
@@ -189,7 +207,13 @@ where
...
-            (Operator(op_pattern), CoreData::Physical(PhysicalOp(UnMaterialized(goal)))) => {
+            (
+                Operator(op_pattern),
+                CoreData::Physical(PhysicalOp {
+                    operator: UnMaterialized(goal),
+                    ..
+                }),
+            ) => {
                 // Yield the goal back to the caller and provide a callback to match expanded value against the pattern.
                 EngineResponse::YieldGoal(
```
yliang412, line 203: Should cost budget propagation also happen here? We probably don't do it in this PR, but in a later one.

optd-dsl/src/analyzer/hir.rs
```diff
@@ -88,14 +88,89 @@ pub struct Operator<T> {
...
+}
+
+impl<T> LogicalOp<T> {
+    /// Creates a new logical operator without a group ID
+    ///
+    /// Used for representing operators that are not yet assigned to a group
+    /// in the optimizer.
+    pub fn logical(operator: Materializable<Operator<T>, GroupId>) -> Self {
+        Self {
+            operator,
```
yliang412, line 103: Be very careful here, as we allow the construction of 

```rust
LogicalOp {
   operator: Unmaterialized(group_id),
   group_id: None
}
```

yliang412 (commented):
I don't have strong objections. The `LogicalOp` / `PhysicalOp` type feels could be polished a little to avoid errors.

optd-dsl/src/analyzer/hir.rs
AlSchlo: Yes. We need to be careful in the analyzer!

AlSchlo: and when we expand, we need to make sure to pass the extra info

optd-core/src/bridge/from_cir.rs
AlSchlo: Me neither, but not importing CoreData::* clutters a lot of the code :shrug:

optd-dsl/src/engine/eval/match.rs
AlSchlo: Probably not in the engine. That seems weird.

yliang412 (approved):
LGTM

PR #51, opened 2025-04-01
(feat) Generics in Functions of DSL

AlSchlo:
## Problem

We need generics in our language to do anything interesting.

Example:
```
fn <K, V> native_map_get(map: {K: V}, key: K): V? = None
```

Generic ADTs is postponed as future work. Since we already have maps, options, and arrays supported, I don't think we need anything extra immediately to express nice rules.

## Summary of changes

Adapted the parser accordingly, and added documentation in `functions.rs`.

Besides, also pushed back `type.rs` in the analyzer (this was old code that will now become relevant again).


yliang412 (approved):
LGTM. I think in a later PR it might make sense to add some failure cases for the parser. the subtype analysis tests are great!

PR #52, opened 2025-04-01
(feat) Analyzer HIR Metadata Setup

AlSchlo:
## Problem

Setup code for the analyzer. The `HIR` gets wrapped with a metadata field.

Illustration:

```rust
/// Metadata that can be attached to expression nodes
///
/// This trait allows for different types of metadata to be attached to
/// expression nodes while maintaining a common interface for access.
pub trait ExprMetadata: Debug + Clone {}

/// Empty metadata implementation for cases where no additional data is needed
pub struct NoMetadata;

/// Combined span and type information for an expression
pub struct TypedSpan {
    /// Source code location.
    pub span: Span,
    /// Inferred type.
    pub ty: Type,
}

/// Expression nodes in the HIR with optional metadata
///
/// The M type parameter allows attaching different kinds of metadata to expressions,
/// such as type information, source spans, or both.
pub struct Expr<M: ExprMetadata = NoMetadata> {
    /// The actual expression node
    pub kind: ExprKind<M>,
    /// Optional metadata for the expression
    pub metadata: M,
}
```

## Summary of changes

Adapted the engine to take EmptyMetadata (no functional changes), all relevant code to review is in HIR.
The critical pass is that the metadata is all we should need to perform type checking and inference.

This PR is dependent on #51

### DIFF
--- a/optd-dsl/src/analyzer/hir.rs
+++ b/optd-dsl/src/analyzer/hir.rs
@@ -15,9 +15,11 @@
 //! unified representation that can be transformed into optimizer-specific
 //! intermediate representations through the bridge modules.
 
-use std::{collections::HashMap, sync::Arc};
-
 use super::context::Context;
+use super::r#type::Type;
+use crate::utils::span::Span;
+use std::fmt::Debug;
+use std::{collections::HashMap, sync::Arc};
 
 /// Unique identifier for variables, functions, types, etc.
 pub type Identifier = String;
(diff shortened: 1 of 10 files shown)


optd-dsl/src/analyzer/hir.rs
```diff
@@ -269,7 +314,11 @@ pub enum UnaryOp {
...
 #[derive(Debug)]
-pub struct HIR {
+pub struct HIR<M: ExprMetadata = NoMetadata> {
     pub context: Context,
     pub annotations: HashMap<Identifier, Vec<Annotation>>,
+    pub expressions: Vec<Expr<M>>,
 }
+
+/// Type alias for HIR with both type and source location information
+pub type TypedSpannedHIR = HIR<TypedSpan>;
```
yliang412, line 324: You might just use `HIR<TypedScan>`, it's even shorter 🤣.

yliang412 (approved):
Looking forward to the analyzer!

PR #53, opened 2025-04-03
(feat) AST->HIR Code prep

AlSchlo:
This PR solves a bunch of problems in the DSL/HIR to allow for the initial transformation from AST -> HIR. Notably:

1. Adds  `$` and `*` type extensions in the parser & type system.

`Type$` stands for Costed(Type) is covariant and inherits Stored(Type)
`Type* `stands for Stored(Type) is covariant and inherits Type.

2. Adds an implementation of map in the HIR and the engine where the keys are allowed to be hashed and compared against each other.

3. Adds `#` to distinguish between field accesses in structs, and member function calls (with `.`).

4. Fixes the `Call` evaluate engine function to include indexing arrays, tuples, maps, and operators (including a group/goal expansion when field access is needed).

5. Added the `Never` type for the Fail expression in the type system.


optd-dsl/src/cli/basic.op
```diff
@@ -93,6 +93,8 @@ data JoinType =
 
 fn <K, V> map_get(map: {K: V}, key: K): V = None
 
+fn empty: V
+
```
connortsui20, line 97: I don't really understand what this syntax is...

optd-dsl/src/analyzer/semantic_checker/error.rs
```diff
@@ -2,10 +2,33 @@ use ariadne::{Report, Source};
...
+            name,
+            first_span,
+            duplicate_span,
+        }
+        .into()
+    }
+}
 
-impl Diagnose for SemanticError {
+impl Diagnose for Box<SemanticError> {
```
connortsui20, line 31: I think now would probably be a good time to invest in using a third-party library like `thiserror` or `snafu` instead of rolling our own error types. Note that it's probably fine if you don't do it now, but it will be much harder to integrate better error types with more code. It's up to you, but personally I feel that those error libraries bring so many QOL features that make development so much nicer

optd-dsl/src/lexer/error.rs
```diff
@@ -16,12 +16,12 @@ pub struct LexerError {
...
     /// Creates a new lexer error from source code and a Chumsky error.
-    pub fn new(src_code: String, error: Simple<char, Span>) -> Self {
-        Self { src_code, error }
+    pub fn new(src_code: String, error: Simple<char, Span>) -> Box<Self> {
+        Self { src_code, error }.into()
     }
 }
 
-impl Diagnose for LexerError {
+impl Diagnose for Box<LexerError> {
```
connortsui20, line 24: Yeah using something like `thiserror` or `snafu` means you don't have to write so much boilerplate like this

connortsui20 (approved):
I don't  have time to look at this in depth, but in general nothing looks out of the ordinary. I do think that error handling methodology could be improved, and from my experience having that done well is super helpful when actually doing the compilation (where a lot of stuff can go wrong)

optd-dsl/src/cli/basic.op
AlSchlo: Oh this will get rejected by the analyzer :-)
It is a Udf without arguments. Hence does not make sense.

optd-dsl/src/analyzer/semantic_checker/error.rs
AlSchlo: Good suggestion for a later refactor IMO. I don't wanna learn another library now.

optd-dsl/src/lexer/error.rs
AlSchlo: I am already using `Ariadne` for error reporting generation in source code. I believe it is somewhat well organized.

PR #54, opened 2025-04-04
(feat) AST -> HIR<TypedSpan> Conversion

AlSchlo:
Implements the first pass of AST->HIR conversion in `analyzer/from_ast`.

In addition to obtaining a `HIR<TypedSpan>`, the conversion process:
1. Desugars syntax.
2. Builds the ADT type registry.
3. Ingests explicit type annotations. 

You may experiment with it as follows:
`cargo run -- compile examples/example.dsl --print-ast --print-hir`


optd-dsl/src/engine/eval/match.rs
```diff
@@ -2,15 +2,15 @@ use crate::{
...
+            CoreData, Expr, LogicalOp, MatchArm, Materializable, Operator, Pattern, PatternKind,
+            PhysicalOp, Value,
         },
     },
     engine::{Continuation, EngineResponse},
 };
 use crate::{capture, engine::Engine};
 use Materializable::*;
-use Pattern::*;
+use PatternKind::*;
```
connortsui20, line 13: Nit: In general exporting enums is usually not good practice because it clutters the namespace. For some things like the `Expr` where there are so many different variants and it is used all over the file it can make sense, but otherwise you should `use` the enums locally inside the functions to make things clear.

optd-dsl/src/analyzer/semantic_checker/error.rs
```diff
@@ -1,10 +1,25 @@
-use ariadne::{Report, Source};
-
 use crate::utils::{error::Diagnose, span::Span};
+use ariadne::{Color, Label, Report, ReportKind, Source};
+
+/// Wrapper for semantic analysis errors
+#[derive(Debug)]
+pub struct SemanticError {
```
connortsui20: I understand that you don't want to learn a new library but just using 1 or 2 of the macros in `thiserror` would prevent a lot of the error boilerplate present throughout this entire crate (not just this file). In the next couple of passes it's likely going to get even more bloated because you'll want to represent different kinds of errors and have to keep modifying your error types to play nicely with each other. [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) is really not that hard to learn (maybe 30 minutes) and `snafu` isn't that much harder. It's up to you, but that is my recommendation

optd-dsl/src/analyzer/hir.rs
```diff
@@ -59,11 +59,18 @@ pub struct TypedSpan {
 }
 impl ExprMetadata for TypedSpan {}
 
+/// User-defined function: either linked or unlinked
+#[derive(Debug, Clone)]
+pub enum UdfKind<M: ExprMetadata> {
+    Linked(fn(Vec<Value<M>>) -> Value<M>),
+    Unlinked(Identifier),
```
connortsui20, line 66: A bit confused by this, I feel like it makes more sense to simply enforce that all UDFs have to have a known static location always: so you could represent a UDF with an identifier or literally just a string and every single usage of that UDF would call into a known static table of function pointers. That way you also don't lose information when you have a linked UDF (there's no way to get the name of the UDF when you have `Linked` variant

connortsui20 (approved):
I don't have time to read through all of the matching code, so I focused on just the type design.

Other than the stuff I mentioned about errors (and on slack about immutable contexts instead of a stack of history for every single context), everything generally looks good to me. I think the UDFs setup is a bit strange but given that you have nothing to work with yet (I still have to write all of the E2E structure for that) I think this is fine to merge and we can fix that later.

optd-dsl/src/analyzer/from_ast/types.rs
```diff
@@ -0,0 +1,247 @@
...
+//! This module contains functions for converting AST type nodes to their
+//! corresponding HIR type representations.
+
+use crate::analyzer::hir::Identifier;
+use crate::analyzer::types::Type;
+use crate::parser::ast;
+use std::collections::HashSet;
+
+/// Converts an AST type to an HIR type.
+pub(crate) fn convert_type(ast_type: &ast::Type, generics: &HashSet<Identifier>) -> Type {
```
yliang412, line 28: Could you add a little note on the purpose of a generics `HashSet`?

yliang412 (approved):
Good work!

optd-dsl/src/engine/eval/match.rs
AlSchlo: That's actually a great suggestion (i.e. importing in function).
I won't do it right now, as it would take a while to change it everywhere (and I have more important things to do).

AlSchlo: I will change that habit and change it as I go.

optd-dsl/src/analyzer/semantic_checker/error.rs
AlSchlo: Perhaps you have not looked into [Ariadne](https://github.com/zesterer/ariadne/tree/main)?

AlSchlo: I took a look at thiserror, and it would just stand in the way of doing something Ariadne allows us to do.

optd-dsl/src/analyzer/hir.rs
AlSchlo: Well, you still gotta link it right?
The input will be: Vec<pointers> + DSL src code.
Output will be linked HIR.

optd-dsl/src/analyzer/from_ast/types.rs
AlSchlo: Yup doing that rn.

optd-dsl/src/analyzer/semantic_checker/error.rs
cjtsui: It's solving a different but parallel problem, these are not mutually exclusive things. Also when I say errors I mean error handling beyond just the DSL compiler, yes ariadne works fine for just the compilation step

optd-dsl/src/analyzer/hir.rs
cjtsui: Right I agree, so why do you need unlinked? It should be easier to just verify that an identifier is either in scope, a function, or a udf than to have it be unlinked and fill that in later right? Unless unlinked has some other purpose

AlSchlo: Ah, so that I don't do all passes at once.
That's the main reason.

I don't want super dense code that's hard to read where we link / check scope / type infer / etc.

AlSchlo: Specifically given that this part of the code is never going to be the performance bottleneck.

PR #56, opened 2025-04-05
(feat) Scope Checker for Analyzer

AlSchlo:
Implements scope checking of the `HIR<TypedSpan>`, so that we get helpful error messages like:

image


connortsui20 (approved):
LGTM

Issue #58, opened 2025-04-05
{  } Scopes in the DSL


AlSchlo:
Perhaps we might want { } to push a new scope.

This would require remembering this in the parser, and then adding an operator node in the HIR.

PR #59, opened 2025-04-06
(feat) Integrate Block Scoping

AlSchlo:
## Problem

Previously, a new block `{ .. }` did not create a new scope where shadowing may happen. This small PR fixes this issue (#58) by adding the feature, hence making the DSL a bit more ergonomic.

## Summary of changes

Adapted the parser, HIR, engine, and analyzer accordingly.
Also refactored the test of `scope_check.rs`, as those were a bit hard to follow.


yliang412 (approved):
LGTM. Do we already have a test for validating a variable in the inner scope will not escape to an outer scope?

AlSchlo:
Added the missing test.

PR #66, opened 2025-04-11
Add top-level `optd` crate + Fix CI + Fix `Cargo.toml`s

connortsui20:
Eventually, we will want `optd` to be an optimizer service. However, we would also like it to be an embeddable library. This PR creates a top-level `optd` crate that will act as the "glue" between all of our components (which right now is just `optd-core` and `optd-dsl`, but more will probably be created). Users should be able to import a single `optd` dependency and have access to all functionality we want to provide (similar to `tokio`). We do _not_ want a datafusion situation where there could be 5 different datafusion crates you need to import.

There is another subtle reason we need this now. We would like to add catalog functionality to `optd`. Both the rule engine and the optimizer need access to the catalog. However, it is not entirely obvious where one would put an implementation of a catalog right now. It doesn't really make sense to put it in `optd-dsl`, but we also can't put it in `optd-core` because `optd-core` depends on `optd-dsl`, which would result in a circular dependency.

By creating a top-level `optd` crate, we can put the definition of a `Catalog` trait in `optd-dsl`, and then add implementations of catalogs in the `optd` crate. This allows for flexibility in catalog implementations while not adding (arguably) irrelevant code to the `optd-dsl` crate.

I removed some unused dependencies and bumped all `Cargo.toml` dependencies to their latest version. However, I updated the `Cargo.lock` with `-Z direct-minimal-versions`. 

This PR also reintroduces the CI checks that we disabled. The `twox-hash` problem still exists when taking a dependency on `iceberg`, so instead of `-Z minimal-versions`, I've changed it to `-Z direct-minimal-versions`.

### DIFF
--- a/optd-dsl/src/catalog.rs
+++ b/optd-dsl/src/catalog.rs
@@ -1,0 +1,14 @@
+use async_trait::async_trait;
+use std::fmt::Debug;
+
+/// An interface for a database catalog that provides CRUD operations over table metadata.
+///
+/// This trait is similar and derived from Apache Iceberg's
+/// [`Catalog`](https://docs.rs/iceberg/latest/iceberg/trait.Catalog.html) Rust trait.
+///
+/// TODO(connor): IMPORTANT!!! Figure out error handling (ideally unify error types).
+/// TODO(connor): THIS TRAIT IS VERY SUBJECT TO CHANGE.
+#[async_trait]
+pub trait Catalog: Debug + Sync + Send {
+    async fn get_table_columns(&self, table_name: &str) -> Result<Vec<String>, String>;
+}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -59,7 +59,9 @@ jobs:
       - name: cargo clippy
         uses: giraffate/clippy-action@v1
         with:
-          reporter: 'github-pr-check'
+          clippy_flags: --all-targets -- Dwarnings
+          reporter: github-pr-check
+          fail_on_error: true
           github_token: ${{ secrets.GITHUB_TOKEN }}
   # # Since this is not a proper library on crates.io we do not need this.
   # semver:
@@ -100,7 +102,7 @@ jobs:
     strategy:
       matrix:
         # Rust 2024 Edition MSRV
-        msrv: ["1.85.0"] 
+        msrv: ["1.85.0"]
     name: ubuntu / ${{ matrix.msrv }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,47 +45,45 @@ jobs:
       # https://github.com/rust-lang/cargo/issues/6669
       - name: cargo test --doc
         run: cargo test --locked --all-features --doc
-  # TODO: Experienced issue with twox-hash-1.6.3, re-enable once fixed.
-  # Failed workflow: https://github.com/cmu-db/optd/actions/runs/13323412234/job/37211945474?pr=26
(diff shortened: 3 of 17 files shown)


connortsui20:
The error handling situation here is really not good. Ideally I think `optd-dsl` exports 1 error type, then `optd-core` can wrap that error type, and then `optd` can wrap that error type again. Libraries like `thiserror` and `snafu` make this quite easy.

optd-dsl/src/catalog.rs
```diff
@@ -0,0 +1,14 @@
...
+use std::fmt::Debug;
+
+/// An interface for a database catalog that provides CRUD operations over table metadata.
+///
+/// This trait is similar and derived from Apache Iceberg's
+/// [`Catalog`](https://docs.rs/iceberg/latest/iceberg/trait.Catalog.html) Rust trait.
+///
+/// TODO(connor): IMPORTANT!!! Figure out error handling (ideally unify error types).
+/// TODO(connor): THIS TRAIT IS VERY SUBJECT TO CHANGE.
+#[async_trait]
```
AlSchlo, line 11: Why don't we use the Catalog defined in `optd`?

AlSchlo: i.e. Shouldn't that trait be in the the root crate?

connortsui20: Then optd core and optd DSL can't access it?

connortsui20:
I think at this point we are running into too many cyclic dependencies and I don't think we need to have this problem. I think there are several things we can consider:

1. Unify `optd-core` and `optd-dsl` into one crate so that they can depend on each other more easily (that way we can add a `Catalog` trait easily into the same crate).
2. Make `optd-core` similar to how `axum` treats [`axum-core`](https://github.com/tokio-rs/axum/tree/main): the core simply being types and definitions that other crates in the workspace can implement. I like this option: since we believe that the design of this optimizer is modular, by defining interfaces and types in one crate, it means that anyone in the future can come along and implement a specific part.
3. Add an `optd-catalog` crate. I am very much not in favor of this, but it is certainly an option.

cc @AlSchlo @SarveshOO7 @yliang412 interested in what you all think

optd-dsl/src/catalog.rs
AlSchlo: Ah, can't you import it?

connortsui20: Import the catalog from from root into dsl? But the root is re-exporting the dsl and so it depends on it as well

AlSchlo: Ah right!

> There is another subtle reason we need this now. We would like to add catalog functionality to optd. Both the rule engine and the optimizer need access to the catalog. However, it is not entirely obvious where one would put an implementation of a catalog right now. It doesn't really make sense to put it in optd-dsl, but we also can't put it in optd-core because optd-core depends on optd-dsl, which would result in a circular dependency.

So this PR doesn't solve that? Just trying to understand if it is a temporary hack.

connortsui20: Solve what specifically? I put the catalog definition in the optd-dsl crate and added an implementation in optd.

If you're talking about it not making sense then no that isn't solved, see the comment below for options.

AlSchlo: Oh missed that one!!!

AlSchlo:
I think putting it all in one crate might be easier all things considered.
At least for the foreseeable future.

Option 2. is cool but kinda looks like YAGNI right now.

AlSchlo (approved):

PR #67, opened 2025-04-11
(feat) Add Return and Propagate Fail Correctly

AlSchlo:
## What & Why

This PR adds a `return` operator in the `HIR` and in the `Engine`, and likewise fixes the behavior of `Fails` **to propagate correctly up the call stack**.

This has been implemented by adding a field to the `Engine`, which keeps track of the upper function call continuation.
Existing test cases have been adapted: (related to `Fail`), and a new one for `Return` has been added.

```rust
     // Define a function:
     // fn test_return(x: I64) = 
     //   let result = if x < 10 {
     //     return "too small"
     //   } else {
     //     "big enough"
     //   }
     //   "result is: " ++ result
     // }
```

### Some Refactoring

- Made all `eval` functions member functions of `Engine`.

### Follow-up Work (less urgent)

- Adding a return keyword in the `AST` and mapping it to the `HIR`.
- Adding the Rust-like `?` operator to unwrap `option` types.
- For some unknown reason I had to make the Messages clonable, and had to make the one shot channel a shared channel. Any idea to avoid that? It doesn't seem like we really need the `Clone` bound, but removing it leads to errors.


yliang412:
> For some unknown reason I had to make the Messages clonable, and had to make the one shot channel a shared channel. Any idea to avoid that? It doesn't seem like we really need the Clone bound, but removing it leads to errors.

I think it's because `Engine<O>` and if we want the engine to be clonable, then you have to make the `continuation` clonable.

Issue #63, opened 2025-04-10
Support Eq if all members are Eq


AlSchlo:
In the engine

PR #71, opened 2025-04-14
(feat) Eq for Value/Value members (#63)

lonless9:
## Problem

The Value type can't auto-derive Eq due to non-comparable function members (e.g., closures). This blocks its usage.

## Summary of changes

Manually implemented PartialEq/Eq for Value and its components.


AlSchlo:
This behavior is not desirable. It does not make sense to eq for example closures, or floats.
We already have a function util that checks if two values are eq (for the subset that is eq), although it is not applied consistently yet -- a refactor might be needed.

PR #77, opened 2025-04-19
Create `compile_hir` compilation entrypoint.

connortsui20:
## Problem

The only place one could compile an HIR was through the `optd-cli` binary. This PR brings this behavior back into the library so we can extend its functionality with UDFs (in a future PR).

## Summary of changes

Created a `Config` struct inside `compile.rs` that largely mimics the previous `clap` struct but cleans it up a bit. This `Config` contains a flattened `Verbosity` struct that groups the different verbosity settings together when compiling. This makes it much easier to pass all of the configurations directly to functions without expanding the fields.

The `compile_hir` function is also located in `compile.rs`, and it largely copies the function that was previously in `optd/src/dsl/cli/main.rs`. This will make it easy to modify what we can pass into compilation (namely a function table for UDFs).

### DIFF
--- a/optd/src/dsl/cli/main.rs
+++ b/optd/src/dsl/cli/main.rs
@@ -6,34 +6,31 @@
 //!
 //! ```
 //! # Compile a DSL file (parse and analyze):
-//! optd compile path/to/file.op
+//! optd-cli compile path/to/file.opt
 //!
 //! # Compile with verbose output:
-//! optd compile path/to/file.op --verbose
+//! optd-cli compile path/to/file.opt --verbose
 //!
 //! # Print intermediate representations:
-//! optd compile path/to/file.op --print-ast --print-typedspan-hir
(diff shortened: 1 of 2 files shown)


AlSchlo (approved):
lgtm

Question: 

> This will make it easy to modify what we can pass into compilation (namely a function table for UDFs).

Could you explain? (why is it easier)

connortsui20:
> > This will make it easy to modify what we can pass into compilation (namely a function table for UDFs).
> 
> Could you explain? (why is it easier)

The user of `optd` has to pass in a table of functions somehow, which means they need to add `optd` as a dependency and be able to compile stuff on their own. If they can't access a `compile_hir` function then they can't manipulate the pipeline in any way, which means they can't add their own UDFs. (By can't manipulate I just mean they can't pass in anything else other than a file path)

PR #78, opened 2025-04-21
Implement UDF linking

connortsui20:
## Summary of changes

This PR implements linking of UDFs by passing in a table of function pointers into `compile_hir`.

This PR also adds the `Catalog` field to both the rule engine and the optimizer as a trait object (`Arc<dyn Catalog>`). I strongly considered making this a generic parameter, but the problem with that is then the `C: Catalog` type parameter would need to be added to literally every structures and function, and I that would significantly reduce readability. The performance is essentially negligible because it is unlikely the compiler could make significant optimizations based on the implementation of the `Catalog`.

Ideally I would add tests that test the functionality of the catalog as well but I don't have time (and also I don't have any reason to think that it wouldn't work now). Once I get an interface that the optimizer needs from the catalog from @SarveshOO7 and @yliang412, I can add those tests easily.

### DIFF
--- a/optd/src/dsl/analyzer/errors.rs
+++ b/optd/src/dsl/analyzer/errors.rs
@@ -275,15 +275,12 @@ impl Diagnose for Box<AnalyzerError> {
                 .with_help("Add at least one parameter or a receiver to this function")
                 .with_note("Functions without parameters are not supported in this language")
                 .finish(),
-            UnknownUdf { name, span } => Report::build(ReportKind::Error, span.clone())
-                .with_message(format!("Unknown UDF found: '{}'", name))
-                .with_label(
-                    Label::new(span.clone())
-                        .with_message("UDF linker error")
-                        .with_color(Color::Magenta),
-                )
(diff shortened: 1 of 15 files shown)


optd/src/dsl/analyzer/errors.rs
```diff
@@ -262,6 +275,15 @@ impl Diagnose for Box<AnalyzerError> {
                 .with_help("Add at least one parameter or a receiver to this function")
                 .with_note("Functions without parameters are not supported in this language")
                 .finish(),
+            UnknownUdf { name, span } => Report::build(ReportKind::Error, span.clone())
```
AlSchlo: I believe you could use the `build_single_span_report` helper.

AlSchlo (approved):
Awesome, thank you so much for doing this!!
All looks good and clean, except perhaps the report generator where you could use the single span util.

optd/src/dsl/analyzer/errors.rs
connortsui20: Makes sense. Could you push onto this PR? I don't have time the rest of today

PR #79, opened 2025-04-21
(feat) Type Inference

AlSchlo:
## Overview

This PR brings in the long awaited type inference to the DSL. This not only makes rules more ergonomic to write, **but it is also needed for correctness in the optimizer**.

The engine needs to know what types are `Logical` or `Physical`, as they behave differently in the `engine`. Thus, all expressions **must** be resolved to a type.

Furthermore, type inference is also used to verify if field accesses and function calls are valid (as these are type-dependent checks).

Some examples:

image

image

image

image

## Strategy 

The type inference works in three phases.

1. In `from_ast`, during the initial `AST` -> `HIR<TypedSpan>` transformation, we create and add all implicit and explicit type information from the program (e.g. Literals like `1` or `"hello"`, function annotations, etc.). For types that are `Unknown`, we generate a new ID and assign the type to either `Descending` (for unknown closure parameters and map keys), or `Ascending` (for all the rest) — more details about these modes in (3).

2. `Constraints` are generated in the `generate.rs` file, which also performs scope-checking now (for convenience, as both code paths were extremely similar). Constraints indicate `subtype` relationship, `field_accesses`, and function `calls`. 

For example:

`let a: Logical = expr`
generates the constraint
`Logical :> typeof(expr)`

3. The last step to resolve the unknown types is to use a solver (credit to @AarryaSaraf for the base algorithm idea). It works as follows:

```
// Pseudocode for the constraint solver
function resolve():
    anyChanged = true
    lastError = null

    // Keep iterating until we reach a fixed point (no more changes)
    while anyChanged:
        anyChanged = false
        lastError = null
        
        // Check each constraint and try to refine unknown types
        for each constraint in constraints:
            result = checkConstraint(constraint)
            
            if result is Ok(changed):
                anyChanged |= changed  // Track if any types were refined
            else if result is Err(error, changed):
                anyChanged |= changed  // Still track changes
                lastError = error      // Remember the last error
        
    // After reaching fixed point, return error if any constraint failed
    return lastError ? Err(lastError) : Ok()
```

During type inference, the method refines unknown types to satisfy subtyping constraints according to their variance:

- When an `UnknownAsc` type is encountered as a parent, it is updated to the least upper bound (LUB) of itself and the child type. These types start at `Nothing` and ascend up the type hierarchy as needed.
- When an `UnknownDesc` type is encountered as a child, it is updated to the greatest lower bound (GLB) of itself and the parent type. These types start at `Universe` and descend down the type hierarchy as needed.
- When an `UnknownAsc` type is encountered as a child, its resolved type is checked against the parent type.
- When an `UnknownDesc` type is encountered as a parent, its resolved type is checked against the child type.

This refinement process happens iteratively until the system reaches a stable state where no more unknown types can be refined. At that point, if any constraints remain unsatisfied, the solver reports the most relevant type error.

The key insight of this algorithm is that it makes monotonic progress - each refinement step either:

- Successfully resolves a constraint
- Refines an unknown type to be more specific
- Identifies a type error

By tracking whether any types changed during each iteration and continuing until we reach a fixed point, we ensure all types are resolved as completely as possible before reporting any errors.

### Limitations

While the algorithm is theoretically correct, it has the following limitations:

1. Its run time appears to be quadratic. This is not a problem for small programs but might become a compilation bottleneck in the future. Excellent heuristics exist to optimize the order in which constraints get applied.

2. As commented in the code, when resolving the constraint:

`UnknownDesc <: UnknownAsc`

We could either dump the left type to `Nothing`, or pump the right type to `Universe`. However, since pumping/dumping types cannot be undone later on (to avoid exponential run-time), it is possible that we over-dump/pump a specific type. A solution would be to ignore these constraints until all other constraints that can be safely applied have run out. This would result in much better empirical type inference.

3. We postpone as future work (#81):
- Map `Concat` as the keys are contra-variant `let map: {Animal : I64} = {Dog : 3} ++ {Cat : 2}` would fail under the current type checker.
- List pattern matching is still broken.
- Generic functions are not yet supported.

All these above points may be solved by adding specific constraints for each of these, like we did for `field_access` and `call`.

4. Some error messages are a bit confusing to understand, although a lot of effort has been done to make them already better (e.g. see examples above). There is no silver bullet here: probably improving the `span` handling from the `parser` is the correct way forward.

## Testing

Given how hard (and bloated!) it is to test each point in isolation, we simply test whether fully written-out programs pass the type checker or not in `solver.rs`.

Error reporting has been tested manually for a variety of programs, and is expected to improve as we start writing rules.

## Future work

**Focus will be put on the final `HIR` compilation process, which needs to correctly encode identified `Logical` / `Physical` types** and reject ambiguous (albeit correctly) inferred types (i.e. `Nothing` or `Universe`). It is indeed better practice to enforce type annotations in these scenarios.

## Note to Reviewers

Don't read the diff, just read the entire `analyzer/types` directory.


optd/src/dsl/analyzer/types/registry.rs
```diff
@@ -34,56 +34,92 @@ pub enum Type {
...
     Concat,     // For types that can be concatenated (String, Array, Map).
     EqHash,     // For types that support equality and hashing.
     Arithmetic, // For types that support arithmetic operations.
 }
 
+/// Represents a type, potentially with span information.
+///
+/// This struct wraps a TypeKind with optional span information, allowing
+/// types to carry source location data when needed.
+pub type Type = OptionalSpanned<TypeKind>;
```
connortsui20, line 70: I understand what this is doing, but sort of similar to a lot of my reviews in the past, the question is why? In my opinion these generic parameters can add confusion and not a lot of benefit. It seems like the only time `OptionalSpanned` is even used is with `TypeKind`. I think it is the same case with the `Value` type from the previous PR, generics don't really make things cleaner.

Really, we want `Span` to be something like this:

```rust
#[derive(Clone, Copy, PartialEq, Eq, Hash)]
pub struct Span {
    /// The range of character positions (start, end) in the source file
    pub range: (usize, usize),
}
```

And have a single global source file (because it's not like we're ever compiling multiple files, and even if we were we should know what those files are statically), which makes `Span` copyable. I think that would improve ergonomics quite a lot.

For the sake of time, we don't have to do anything, but I think this is worth an issue.

connortsui20 (approved):
I don't have time to go through the whole algorithm so just doing a quick look over some of the code and tests everything looks fine. I have other concerns about Rust-related things but we have more important things to worry about.

PR #83, opened 2025-04-24
Memory memo

connortsui20:
## Problem

Adds an in-memory implementation of the optimizer's memo table.

NOTE: Even though I am opening this PR, this is really authored by @SarveshOO7. I'll be the one reviewing it.


optd/src/core/memo/mod.rs
```diff
@@ -20,24 +35,27 @@ pub enum Status {
...
-    /// ID of the merged group
-    pub group_id: GroupId,
-
-    /// All logical expressions in this group
-    pub expressions: Vec<LogicalExpressionId>,
-}
-
 /// Result of merging two groups.
 #[derive(Debug)]
 pub struct MergeGroupResult {
```
connortsui20: Nit: This is not a great name. Usually if you have "Result" at the end of your type name, it should behave like an actual `Result<T, E>`. Maybe use `MergeGroupInfo` or `GroupMergeInfo` or `GroupMergeData`

optd/src/core/memo/mod.rs
```diff
@@ -20,24 +35,27 @@ pub enum Status {
...
+    pub merged_groups: HashMap<GroupId, Vec<LogicalExpressionId>>,
+}
+
+impl MergeGroupResult {
+    /// Creates a new MergeGroupResult instance.
+    ///
+    /// # Parameters
+    /// * `merged_groups` - Groups that were merged along with their expressions.
+    /// * `new_repr_group_id` - ID of the new representative group id.
+    pub fn new(new_repr_group_id: GroupId) -> Self {
```
connortsui20: This seems to be stale

connortsui20: In fact I would prefer to not have any constructor here as both of the fields of `MergeGroupResult` (which I still think should have a different name) are `pub`, and the only place this `new` function is called would probably be cleaner without `new`.

optd/src/core/memo/mod.rs
```diff
@@ -57,29 +75,60 @@ pub struct MergedGoalInfo {
...
+    pub best_expr: Option<(PhysicalExpressionId, Cost)>,
 
     /// ID of the new representative goal id.
     pub new_repr_goal_id: GoalId,
 }
 
-/// Results of merge operations with newly dirtied expressions.
+/// Result of merging two cost expressions.
 #[derive(Debug)]
+pub struct MergePhysicalExprResult {
```
connortsui20: Maybe rename:

- `MergedGoalInfo` -> `MergedGoal`
- `MergeGoalResult` -> `GoalMergeData` / `GoalMergeInfo`

optd/src/core/memo/mod.rs
```diff
@@ -57,29 +75,60 @@ pub struct MergedGoalInfo {
...
+    // pub dirty_transformations: Vec<(LogicalExpressionId, TransformationRule)>,
+
+    // /// Implementations that were marked as dirty and need new application.
+    // pub dirty_implementations: Vec<(LogicalExpressionId, GoalId, ImplementationRule)>,
 
-    /// Implementations that were marked as dirty and need new application.
-    pub dirty_implementations: Vec<(LogicalExpressionId, GoalId, ImplementationRule)>,
+    // /// Costings that were marked as dirty and need recomputation.
+    // pub dirty_costings: Vec<PhysicalExpressionId>,
+}
```
connortsui20: Dead code

optd/src/core/memo/mod.rs
```diff
@@ -57,29 +75,60 @@ pub struct MergedGoalInfo {
...
 
-    /// Costings that were marked as dirty and need recomputation.
-    pub dirty_costings: Vec<PhysicalExpressionId>,
+pub struct ForwardResult {
+    pub physical_expr_id: PhysicalExpressionId,
+    pub best_cost: Cost,
+    pub goals_forwarded: HashSet<GoalId>,
+}
```
connortsui20: Documentation needed here, what is this (and why is it called Result)?

optd/src/core/memo/mod.rs
```diff
@@ -101,7 +150,24 @@ pub trait Memoize: Send + Sync + 'static {
     ///
     /// # Returns
     /// The properties associated with the group or an error if not found.
-    async fn get_logical_properties(&self, group_id: GroupId) -> MemoizeResult<LogicalProperties>;
+    async fn get_logical_properties(
```
connortsui20: Comment for a few lines above this one (because I couldn't add the review on GitHub):

```rust
#[trait_variant::make(Send)]
pub trait Memoize: Send + Sync + 'static {
```

Why does the memo table need to be all of `Send + Sync + 'static` bounds? This needs documentation for why those are required. _Are_ they even required? `Sync` makes sense to me but the others two don't really make sense.

optd/src/core/memo/mod.rs
```diff
@@ -101,7 +150,24 @@ pub trait Memoize: Send + Sync + 'static {
...
+        &self,
+        group_id: GroupId,
+    ) -> MemoizeResult<Option<LogicalProperties>>;
```
connortsui20: Why does this return an `Option<LogicalProperties>`?

This led me to go look at `GroupState` in `memory.rs` , and the double `Option` is somewhat questionable:

```rust
struct GroupState {
    /// The logical properties of the group, might be `None` if it hasn't been derived yet.
    properties: Option<LogicalProperties>,
    logical_exprs: HashSet<LogicalExpressionId>,
    goals: HashSet<GoalId>,
}

#[derive(Debug, Clone, PartialEq)]
pub struct LogicalProperties(pub Option<PropertiesData>);
```

optd/src/core/memo/mod.rs
```diff
@@ -115,6 +181,9 @@ pub trait Memoize: Send + Sync + 'static {
         group_id: GroupId,
     ) -> MemoizeResult<Vec<LogicalExpressionId>>;
 
+    /// Gets any logical expression ID in a group.
+    async fn get_any_logical_expr(&self, group_id: GroupId) -> MemoizeResult<LogicalExpressionId>;
```
connortsui20: At first glance, this seems like a super strange method to require (it's not obvious why we need this, and there's no documentation on why this is here). It might be good to think about requiring a `get_all_logical_exprs` that returns an **iterator** or even stream of logical expression IDs in any order so that we don't need to have these weird methods.

optd/src/core/memo/mod.rs
```diff
@@ -115,6 +181,9 @@ pub trait Memoize: Send + Sync + 'static {
...
+
     /// Finds group containing a logical expression ID, if it exists.
```
connortsui20: I'm curious (not really part of the review), in what situation is a `LogicalExpressionId` not going to exist? I'm assuming that `LogicalExpressionId` can only be created from a real logical expression (unless this is wrong and we can just have invalid logical expression IDs floating around)? Can logical expressions be destroyed somehow?

optd/src/core/memo/mod.rs
```diff
@@ -138,7 +207,6 @@ pub trait Memoize: Send + Sync + 'static {
     async fn create_group(
         &mut self,
         logical_expr_id: LogicalExpressionId,
```
connortsui20: The doc parameters are stale

optd/src/core/memo/mod.rs
```diff
@@ -156,7 +224,7 @@ pub trait Memoize: Send + Sync + 'static {
         &mut self,
         group_id_1: GroupId,
         group_id_2: GroupId,
-    ) -> MemoizeResult<MergeResult>;
+    ) -> MemoizeResult<Option<MergeResult>>;
```
connortsui20: In what situation does merge group fail? I thought that the whole point what we've been talking about over the past month was to prevent group merge failures

connortsui20 (commented):
This `Memozie` trait is MASSIVE. I think it is worth considering if we should split the `Memoize` trait into several smaller subtraits (divided similarly to how there are literal section in the trait definition itself). So for example have a trait `LogicalMemo` that has all of the methods related to logical properties, expressions, groups. Then have another trait `PhysicalMemo` that has all the physical expressions and goal stuff. Then `RuleStatusMemo`, etc.

Additionally, I am worried about how the integration with the optimizer is going to go because there are several questionable types and signatures that I found in this first pass in just the `Memoize` trait alone. I know that we are very low on time, but I think it is critical that @SarveshOO7 and @yliang412 take just a little bit of time (like 30 minutes) to read through the `Memoize` trait and confirm that the Memo table API makes sense for the things they need in the optimizer (and document why things need to be there). Let me know if you need help with any of this.

optd/src/core/memo/mod.rs
AlSchlo: IIRC (@yliang412 ) can confirm, it was to deal with the case were we retrieve the properties of a group? Anyways, I agree it should be removed.

optd/src/core/memo/mod.rs
AlSchlo: > LogicalExpressionId can only be created from a real logical expression 

Yes.

They cannot be destroyed either.

optd/src/core/memo/mod.rs
```diff
@@ -1,14 +1,29 @@
...
+
+mod memory;
+mod merge_repr;
 
 /// Type alias for results returned by Memoize trait methods
-pub type MemoizeResult<T> = Result<T, Error>;
+pub type MemoizeResult<T> = Result<T, MemoizeError>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MemoizeError {
```
AlSchlo: Why have these as errors rather than having the functions return an option?
If these errors should never happen unless code is wrong, they should not be errors.

optd/src/core/memo/mod.rs
AlSchlo: I would call it `MergeGroupDiff` (but think the current name is OK)

optd/src/core/memo/mod.rs
AlSchlo: oh yeah please get rid of this useless constructor.
That's just code bloat.

optd/src/core/memo/mod.rs
```diff
@@ -57,29 +75,60 @@ pub struct MergedGoalInfo {
...
     pub group_merges: Vec<MergeGroupResult>,
 
     /// Goal merge results.
     pub goal_merges: Vec<MergeGoalResult>,
 
-    /// Transformations that were marked as dirty and need new application.
-    pub dirty_transformations: Vec<(LogicalExpressionId, TransformationRule)>,
+    /// Physical expression merge results.
+    pub physical_expr_merges: Vec<MergePhysicalExprResult>,
+    // /// Transformations that were marked as dirty and need new application.
```
AlSchlo: don't keep dead comments in PR, or create an issue that point to it (e.g. TODO(#xx)) and recaps what needs to be done

optd/src/core/memo/mod.rs
AlSchlo: I myself am confused and do not remember either :+1:

AlSchlo: What does forwarding a goal mean?

AlSchlo: Also no need to add constructors

optd/src/core/memo/mod.rs
AlSchlo: Because of tokio...
The memo lies in the optimizer state, which is sent around co-routines.

optd/src/core/memo/mod.rs
AlSchlo: :+1: should *not* be an option

optd/src/core/memo/mod.rs
connortsui20: Oh in that case then this is part of the review. It would be good to encode these invariants in the types, or if that's too much work at least document this somewhere.

optd/src/core/memo/memory.rs
```diff
@@ -0,0 +1,1111 @@
...
+    }
+}
+
+impl Memoize for MemoryMemo {
+    async fn merge_groups(
+        &mut self,
+        group_id_1: GroupId,
+        group_id_2: GroupId,
+    ) -> MemoizeResult<Option<MergeResult>> {
+        self.merge_groups_helper(group_id_1, group_id_2).await
```
AlSchlo: why have a helper that seems to be the entire implementation?

optd/src/core/memo/memory.rs
```diff
@@ -0,0 +1,1111 @@
...
+    async fn set_logical_properties(
+        &mut self,
+        group_id: GroupId,
+        props: LogicalProperties,
+    ) -> MemoizeResult<()> {
+        let group_id = self.find_repr_group(group_id).await?;
+        let group = self
+            .groups
+            .get_mut(&group_id)
+            .ok_or(MemoizeError::GroupNotFound(group_id))?;
```
AlSchlo: Errors should be the exception, not the correct code path.

AlSchlo (commented):
I would like to have a description of the merge algorithm.
I don't want to spend a day trying to reverse engineer how this works.

Also I don't see tests in this code?

optd/src/core/memo/mod.rs
SarveshOO7: It's dead right now because it hasn't been implemented yet

connortsui20: Then something that is not implemented should not be here... Otherwise make the surrounding code and add `unimplemented!()`

PR #94, opened 2025-05-01
rename types

connortsui20:
## Problem

Some of the names in `mod.rs` are super confusing.

## Summary of changes

Changes the names of several of the types. Look at `memo/mod.rs` for the changes, everything else is just autorenamed.


connortsui20:
@SarveshOO7 @yliang412 This would be a good time to add documentation to `ForwardResult` (now `PropagateCost`) and some of the other fields. Especially if you think that some of these names aren't good - documentation would have helped me decide better names.

Also note that the only reason I'm making this PR into the existing PR is so that we don't have to deal with merge conflicts later.

AlSchlo (approved):

PR #95, opened 2025-05-01
`optd-cli` crate

connortsui20:
## Problem

If we want users of the DSL to add their own UDFs, it makes more sense to have them recompile just the CLI with their functions rather than the entirety of `optd`.

## Summary of changes

Extracts the `cli` module into its own crate.

Ideally we figure out how to let users of the CLI add their own UDFs in an ergonomic way without having to fork our repository. This should be easy (could either use a proc macro or expose more of the compilation functions as building blocks in a library), and can be left to a later point in time.


connortsui20:
whoops forgot to pull in changes

AlSchlo (approved):

PR #97, opened 2025-05-02
Break up Memo trait

connortsui20:
## Problem

The memo trait was way too large.

## Summary of changes

This PR breaks up the Memo table into 4 subtraits. See `src/memo/traits.rs` for more info.


AlSchlo:
Work integrated into #101

PR #101, opened 2025-05-07
(feat) Optimizer & Memo Full Logical -> Logical

AlSchlo:
Implements the in-memory memo for logical -> logical, as well as the optimizer task graph, and the support of `*` stored expressions in the compiler.

All features are tested accordingly, except the optimizer itself which will be postponed in a future PR.


connortsui20:
Not a full review, but if you run `cargo clippy` it will show a bunch of things that you can clean up (other than the things that don't compile)

optd/src/memo/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+pub mod memory;
+
+pub(crate) use error::MemoError;
+pub use memory::MemoryMemo;
+
+/// Result of merging two groups.
+#[derive(Debug)]
+pub struct MergeGroupProduct {
+    /// ID of the new representative group id.
+    pub new_repr_group_id: GroupId,
```
connortsui20: Nit: Don't call this `new` since (if I understand correctly) the representative group should probably already exist or could even have already been a representative group?

optd/src/memo/error.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    /// A [`GoalId`] does not exist in the memo.
+    GoalNotFound(GoalId),
+
+    /// A [`LogicalExpressionId`] does not exist in the memo.
+    LogicalExprNotFound(LogicalExpressionId),
+
+    /// A [`PhysicalExpressionId`] does not exist in the memo.
+    PhysicalExprNotFound(PhysicalExpressionId),
+}
```
connortsui20: As per the clippy warning, all of these error variants are about groups or goals not being found. I am wondering in what scenario these errors will be raised? I believe that the IDs should only really exist if the corresponding object exists in the table, otherwise I think we have a problem. This is only not true if any of these objects (groups, goals, exprs) can be deleted without the caller / task graph knowing about it.

If it is the case that the memo table should never fail (which I think might actually be the case), then it is actually reasonable for methods to panic on broken invariants. Basically, if there are invariants that the Memo table upholds, and the only kind of error is a breaking of said invariants, then we actually don't really need to return errors and we should just panic.

optd/src/memo/memory/mod.rs
```diff
@@ -0,0 +1,608 @@
...
+
+    /// The shared next unique id to be used for goals, groups, logical expressions, and physical expressions.
+    next_shared_id: i64,
+
+    /// Representatives of groups, goals, and expression ids, so that we can process old IDs.
+    repr_group_id: UnionFind<GroupId>,
+    repr_goal_id: UnionFind<GoalId>,
+    repr_logical_expr_id: UnionFind<LogicalExpressionId>,
+    repr_physical_expr_id: UnionFind<PhysicalExpressionId>,
+}
```
connortsui20, line 59: It might be good to break this up into multiple structs and then compose all of them in the `MemoryMemo`? 

For example, the possible structs here could be `GroupInfo`, `GoalInfo`, `(Logical)ExpressionInfo`, `(Logical)ExpressionIndexes`, `MemoRepresentatives` (roughly in order of how they appear in the current struct). `GroupInfo` conflicts but that could just be renamed to `Group` imo.

I think that might help with the implementation below as well because you can encapsulate stuff better.

connortsui20 (changes requested):
Not a full review, but some small things to think about for now. I'll come back to this later this afternoon.

optd/src/memo/memory/helpers.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Gets the set of logical expression IDs that reference a group,
+    /// mapped to their representatives.
+    fn get_referencing_expr_set(&mut self, group_id: GroupId) -> HashSet<LogicalExpressionId> {
+        self.group_referencing_exprs_index
+            .remove(&group_id)
+            .unwrap_or_default()
+            .iter()
+            .map(|id| self.repr_logical_expr_id.find(id))
+            .collect()
+    }
```
connortsui20: The description and method name doesn't seem to correlate with the body, why is this removing / mutating an index if it just need to retrieve something?

optd/src/memo/mod.rs
connortsui20: Ah I see in the other code that it _is_ actually a new group ID.

optd/src/memo/memory/helpers.rs
```diff
@@ -0,0 +1,299 @@
...
+    },
+};
+use std::collections::{HashMap, HashSet};
+
+impl MemoryMemo {
+    pub(super) fn next_shared_id(&mut self) -> i64 {
+        let id = self.next_shared_id;
+        self.next_shared_id += 1;
+        id
+    }
```
connortsui20: This might as well return a `GroupId` (to hide the fact that it is an `i64`)

optd/src/memo/memory/helpers.rs
```diff
@@ -0,0 +1,299 @@
...
+            .get(&group_id_1)
+            .ok_or(MemoError::GroupNotFound(group_id_1))?;
+        let group_info_2 = self
+            .group_info
+            .get(&group_id_2)
+            .ok_or(MemoError::GroupNotFound(group_id_2))?;
+
+        // Verify logical properties match (should be the case for equivalent groups)
+        // Unless there is a bug in the rules.
+        debug_assert!(group_info_1.logical_properties == group_info_2.logical_properties);
```
connortsui20: This should be an `assert!` if it is possible for a bug in the rules to trigger this.

optd/src/memo/memory/helpers.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        // Combine expressions from both groups.
+        let expr_ids_1 = self.get_all_logical_exprs(group_id_1).await?;
+        let expr_ids_2 = self.get_all_logical_exprs(group_id_2).await?;
+        let mut all_expr_ids = HashSet::with_capacity(expr_ids_1.len() + expr_ids_2.len());
+        all_expr_ids.extend(expr_ids_1.iter().copied());
+        all_expr_ids.extend(expr_ids_2.iter().copied());
```
connortsui20: [`union`](https://doc.rust-lang.org/std/collections/hash_set/struct.HashSet.html#method.union) would be more idiomatic and readable

optd/src/memo/error.rs
connortsui20: Looking at the code, I now see that it is possible for `GroupId`s to be removed after merging which is totally fine. However, I still wonder if it makes sense to panic on one of these errors. What would the optimizer do if this happened? I can't really think of anything that the optimizer could do to "recover" if one of these errors was raised.

optd/src/memo/memory/materialize.rs
```diff
@@ -0,0 +1,137 @@
...
+            .children
+            .iter()
+            .flat_map(|child| match child {
+                Singleton(group_id) => vec![*group_id],
+                VarLength(group_ids) => group_ids.clone(),
+            })
+            .for_each(|group_id| {
+                self.group_referencing_exprs_index
+                    .entry(group_id)
+                    .or_insert_with(HashSet::new)
```
connortsui20: clippy warning

optd/src/memo/memory/materialize.rs
```diff
@@ -0,0 +1,137 @@
...
+    async fn materialize_physical_expr(
+        &self,
+        _physical_expr_id: PhysicalExpressionId,
+    ) -> MemoResult<PhysicalExpression> {
+        todo!()
+    }
+}
+
+// Remapping helpers to get canonical expressions, goals, etc.
+impl MemoryMemo {
```
connortsui20: Might as well put this in the helpers module

optd/src/memo/memory/mod.rs
```diff
@@ -0,0 +1,608 @@
...
+                    .await?;
+
+                pending_merges.extend(new_pending_merges);
+            }
+        }
+
+        // Consolidate the merge results by replacing the incremental merges
+        // with consolidated results that show the full picture.
+        self.consolidate_merge_results(merge_operations).await
+    }
```
connortsui20: I'm a bit confused by this, the algorithm description says that it maintains a queue of group pairs that need to be merged, but this code seems like it does these in batches?

If I understand this correctly, this would do the exact same thing:

```rust
        let mut merge_operations = vec![];
        let mut pending_merges = VecDeque::from(vec![(group_id_1, group_id_2)]);

        while let Some((g1, g2)) = pending_merges.pop_front() {
            // <-- The exact same code under the for loop -->
            // ...

            pending_merges.extend(new_pending_merges);
        }

        // Consolidate the merge results by replacing the incremental merges
        // with consolidated results that show the full picture.
        self.consolidate_merge_results(merge_operations).await
```

connortsui20 (changes requested):
Very nice job! I like the design here and the fact that generally different ideas are separated along different methods and types. I think the new group ID for newly merged groups is also generally a good idea now - before when we were trying to think about parallelism it wasn't ideal but since that is out the window we don't really care about that anymore.

I suggested a few things, but the only thing that I guess is worth discussing is the error handling on the memo table. Basically, what should happen if something goes wrong? See my comments above.

optd/src/memo/memory/helpers.rs
AlSchlo: Because path compression of union-find I think.

connortsui20: That doesn't really make sense... That remove is for a hash maps that is separate from the union find stuff

optd/src/memo/memory/helpers.rs
connortsui20: See #104

optd/src/memo/memory/materialize.rs
connortsui20: See #104

optd/src/memo/memory/mod.rs
AlSchlo: Yeah I agree, but these are the types of things I do when I get bored and want it to look nice.
There's more pressing matters.

optd/src/memo/memory/mod.rs
AlSchlo: will change now

AlSchlo: i struggled a bit with the borrow checker and this is what gen AI gave me ... sigh

optd/src/optimizer/memo_io/egest.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// # Parameters
+    /// * `expression_id` - ID of the physical expression to transform into a complete plan.
+    ///
+    /// # Returns
+    /// * `Ok(Some(PhysicalPlan))` if all children plans were successfully constructed from their IDs.
+    /// * `Ok(None)` if any goal ID lacks a best expression ID.
+    /// * `Err(Error)` if a memo operation fails.
+    #[async_recursion]
+    #[allow(dead_code)]
+    pub(crate) async fn egest_best_plan(
```
connortsui20, line 25: Maybe don't call this `egest_best_plan` since it is just materializing a physical plan (not necessarily the best)?

optd/src/optimizer/memo_io/egest.rs
connortsui20, line 1: Generally pretty clean, though there's some duplicate behavior that could probably be packaged into a helper method that returns a materialized expression and the materialized children, though that's just a nit.

connortsui20 (approved):
LGTM! I only looked at the things related to the memo but I think this is fine to merge and clean up later.

optd/src/optimizer/memo_io/egest.rs
AlSchlo: Materialize is more vocabulary for making a PhysicalExpressionId to a PhysicalExpression.

optd/src/optimizer/memo_io/egest.rs
AlSchlo: Yeah tried that a while ago, but the generics were to bizarre for humans to comprehend.

optd/src/memo/memory/helpers.rs
AlSchlo: renamed to take

PR #103, opened 2025-05-08
Memory memo tests

jopereira:
## Problem

The memo is being tested only through optimizer tests.

## Summary of changes

Add simple tests for key memo functionality, including insert, lookup, merge, and recursive merge of groups.

Limitation: These tests are currently using only logical expressions.

### DIFF
--- a/optd/src/memo/memory.rs
+++ b/optd/src/memo/memory.rs
@@ -1131,3 +1131,77 @@ impl MemoryMemo {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::memo::tests::*;
+    use super::*;
+
+    #[tokio::test]
+    async fn test_lookup_same() {
+        let mut memo = MemoryMemo::default();
+
+        let g0 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+        let g1 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+
+        assert_eq!(g0, g1);
+    }
+
+    #[tokio::test]
+    async fn test_lookup_different() {
+        let mut memo = MemoryMemo::default();
+
+        let g0 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+        let g1 = lookup_or_insert(&mut memo, 1, Vec::new()).await;
+
+        assert_ne!(g0, g1);
+    }
+
+    #[tokio::test]
+    async fn test_add_to_group() {
+        let mut memo = MemoryMemo::default();
+
+        let g0 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+        insert_into_group(&mut memo, 1, Vec::new(), g0).await;
+
+        assert_eq!(retrieve(&memo, g0).await, vec![0, 1]);
+    }
+
+    #[tokio::test]
+    async fn test_merge() {
+        let mut memo = MemoryMemo::default();
+
+        let g0 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+        insert_into_group(&mut memo, 1, Vec::new(), g0).await;
+        let g2 = lookup_or_insert(&mut memo, 2, Vec::new()).await;
+        insert_into_group(&mut memo, 3, Vec::new(), g2).await;
+        let g4 = insert_into_group(&mut memo, 0, Vec::new(), g2).await;
+
+        assert_eq!(retrieve(&memo, g4).await, vec![0, 1, 2, 3]);
+    }
+
+    #[tokio::test]
+    async fn test_recursive_merge() {
+        let mut memo = MemoryMemo::default();
+
+        let g0 = lookup_or_insert(&mut memo, 0, Vec::new()).await;
+        let g1 = insert_into_group(&mut memo, 1, Vec::new(), g0).await;
+        let g4 = lookup_or_insert(&mut memo, 4, vec![g1]).await;
+        insert_into_group(&mut memo, 5, Vec::new(), g4).await;
+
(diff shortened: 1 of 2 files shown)


AlSchlo:
Thanks for testing this! @jopereira 

I have integrated the tests into #101, which is a branch I made where I rewrote the memo and optimizer.
The memo implementation should be faster & much more readable now.

Could you please try your fuzzing based tests again?

I plan to merge that PR once I wrap up merging the logical task graph (which should take me one extra day + testing).

jopereira:
> Could you please try your fuzzing based tests again?

There seems to be a problem... To reproduce:

    git clone https://github.com/jopereira/memobench.git
    cd memobench
    cargo run --features=optd_mem -- -d -g 100 -e 10 -a -r --shuffle=merge --seed 13002543991537748300 optd-mem

This results in the same expression twice in the same group:

    assertion `left == right` failed: incorrect memo
      left: [174, 175, 176, 177, 178, 178, 179]
     right: [174, 175, 176, 177, 178, 179]

The sequence of operations leading to this are:

- 174, 175 (creates first group)
- 176, 177 (creates second group)
- 178, 179 (greates third group)
- 174, 176 (triggers merge first+second)
- 178, 174 (triggers merge with third) -> problem here, 178 is recognized as the same to merge groups, but kept twice!

The full sequence of operations can be obtained with the --raw-csv option. 

It is possible that I'm not using the new API correctly. Finally, this doesn't seem to be fully deterministic... which is surprising me as the test should be fully determined by the seed.

AlSchlo:
Thanks, I'll take a look tomorrow. 
The lack of determinism is not surprising, given that we used HashSets (even a static seed will not make it deterministic).

AlSchlo:
```rust
    #[tokio::test]
    async fn test_duplicate_expr_after_merge() {
        let mut memo = MemoryMemo::default();

        // 1. Create first group with expressions 174, 175
        let g1 = lookup_or_insert(&mut memo, 174, vec![]).await;
        insert_into_group(&mut memo, 175, vec![], g1).await;

        // 2. Create second group with expressions 176, 177
        let g2 = lookup_or_insert(&mut memo, 176, vec![]).await;
        insert_into_group(&mut memo, 177, vec![], g2).await;

        // 3. Create third group with expressions 178, 179
        let g3 = lookup_or_insert(&mut memo, 178, vec![]).await;
        insert_into_group(&mut memo, 179, vec![], g3).await;

        // Verify initial state
        assert_eq!(retrieve(&memo, g1).await, vec![174, 175]);
        assert_eq!(retrieve(&memo, g2).await, vec![176, 177]);
        assert_eq!(retrieve(&memo, g3).await, vec![178, 179]);

        // 4. Merge first and second groups (174, 176)
        memo.merge_groups(g1, g2).await.unwrap();

        // Get the representative of the merged group
        let merged_g1_g2 = memo.find_repr_group_id(g1).await.unwrap();

        // Verify merged state
        let merged_exprs = retrieve(&memo, merged_g1_g2).await;
        assert_eq!(merged_exprs, vec![174, 175, 176, 177]);

        // 5. Merge third group with merged group (178, 174)
        memo.merge_groups(g3, merged_g1_g2).await.unwrap();

        // Find the final representative group
        let final_group = memo.find_repr_group_id(g3).await.unwrap();

        // Retrieve expressions in the final group
        let final_exprs = retrieve(&memo, final_group).await;

        // This is where the bug manifests - expression 178 appears twice
        // The assert will fail if the bug is present
        assert_eq!(
            final_exprs,
            vec![174, 175, 176, 177, 178, 179],
            "Merged group should contain exactly one copy of each expression"
        );
    }
    ```
    
    Just added this unit test and it works as intended. Will look a bit more into the benchmark implementation.

**Comment by AlSchlo** [comment]
EDIT: I understood how to interpret the raw csv file. Will replicate properly.

**Comment by AlSchlo** [comment]
@jopereira I was able to solve the problem by adding a few lines here (add repr mapping):
```rust
    async fn get_all_logical_exprs(
        &self,
        group_id: GroupId,
    ) -> MemoResult<HashSet<LogicalExpressionId>> {
        let group_id = self.find_repr_group_id(group_id).await?;
        Ok(self
            .group_info
            .get(&group_id)
            .ok_or(MemoError::GroupNotFound(group_id))?
            .expressions
            .iter()
            .map(|id| self.repr_logical_expr_id.find(id))
            .collect())
    }
```
That is the symptom, and seems to solve it. However, it is still very unclear to me why we need that. As I thought my memo had expressions representatives as an invariant.

AlSchlo:
I benchmarked against optd-original too, and the implementation is orders of magnitudes more efficient :D

AlSchlo:
RESOLVED: @jopereira 

[Commit that fixes it](https://github.com/cmu-db/optd/pull/101/commits/bb5ee0249e709dd55c36a12afb852bb14bc609df)

The `create_group` is for good measure, the issue was during the replacement of merged expressions. I was adding the new one too many times and violating the invariant of "a logical expr can map to at most one group". Not a critical bug, but a bug anyway.

You should be able to fuzz it again. I think it's all good. Would be interesting to see benchmarks too, since I cannot run Calcite nor Redis (no Java).

Thanks!

I benchmarked against optd-original at least, and the new implementation is *orders* of magnitudes more efficient. It seemed to do something sequential.

jopereira:
Good! Indeed, I cannot make it crash now.

I'll run a full set of benchmarks.

AlSchlo:
Probably a good idea to try with a [different hashtable implementable](https://docs.rs/hashbrown/latest/hashbrown/) (given how much we use those).

AlSchlo:
I have replaced it with the SwissTable implementation. It's always significantly better from my experience.

PR #108, opened 2025-05-14
Some memo refactors

connortsui20:
## Problem

There were a few inefficiencies in the memo table (mainly extraneous clones, and how remapping of objects like goals and expressions work). Also, some of the function docs were quite verbose.

## Summary of changes

This PR cleans up some of the memo table modules style and documentation, as well as removes several extraneous clones.

### DIFF
--- a/optd/src/memo/memory/helpers.rs
+++ b/optd/src/memo/memory/helpers.rs
@@ -34,15 +34,26 @@ impl MemoryMemo {
         PhysicalExpressionId(id)
     }
 
-    /// Takes the set of [`LogicalExpressionId`] that reference a group, mapped to their
-    /// representatives.
-    fn take_referencing_expr_set(&mut self, group_id: GroupId) -> HashSet<LogicalExpressionId> {
+    /// Merges the two sets of logical expressions that reference the two groups into a single set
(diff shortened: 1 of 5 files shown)


optd/src/memo/memory/helpers.rs
```diff
@@ -214,55 +242,53 @@ impl MemoryMemoHelpers for MemoryMemo {
...
 
-            // Only add the new expression if it is not already in the group.
+            // Only add the new expression to the old expression's group if it does not exist in any
+            // group yet.
+            // TODO(connor) make sense of these docs.
+            // Since we are effectively adding an identical expression into this group, we
+            // technically will have 2
             // We do not want to add the same logical expression to two groups.
             // The merge will handle that, if it is needed.
             if new_group_opt.is_none() {
```
connortsui20: I don't really understand this comment, basically not sure why adding an expression to two groups is relevant here. Is this supposed to be explaining how we might be adding an equivalent expression to the same group, and this will be "cleaned up" during merging?

AlSchlo (approved):
Looks all good to me, please ensure that nothing gets broken by running the fuzzing tests again.

PR #111, opened 2025-05-22
Add fuzz testing for memo

jopereira:
## Problem

The memo data structure is too complex to come up with an exhaustive set of tests.

## Summary of changes

Add fuzz testing based on the data generator and workload from https://github.com/jopereira/memobench. For now, this tests only logical expressions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2020,6 +2020,7 @@ dependencies = [
  "iceberg-catalog-memory",
  "once_cell",
  "ordered-float 5.0.0",
+ "rand 0.8.5",
  "tempfile",
  "tokio",
  "trait-variant",
--- a/optd/Cargo.toml
+++ b/optd/Cargo.toml
@@ -28,3 +28,4 @@ tokio = { version = "1.45.0", features = [
     "rt-multi-thread",
     "test-util"
 ] }
+rand = "0.8.5"
--- a/optd/src/memo/fuzz.rs
+++ b/optd/src/memo/fuzz.rs
@@ -1,0 +1,236 @@
+
+use rand::rngs::StdRng;
+use rand::distributions::WeightedIndex;
+use rand::prelude::*;
+use std::cmp::max;
+use std::collections::HashSet;
+use std::time::Instant;
+use crate::cir::{LogicalExpression, LogicalProperties, OperatorData, GroupId, Child};
+use crate::memo::{Memo,Materialize};
+
+#[derive(Clone)]
+pub struct FuzzExpression {
+    pub op: usize,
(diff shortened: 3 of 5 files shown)


AlSchlo:
I can't find the `FuzzData`, etc. structs. Could you please make the branch in `cmu-db/optd`?

jopereira:
Missed a file. It should be there now.

AlSchlo:
It seems that the fuzzing code only creates one group that contains more than 1 expression.

Troubleshooting right now.

jopereira:
The default arguments are low. Try 10k groups and 100 expressions. It will however take some time to run and probably should be disabled by default.

AlSchlo:
Indeed, for some reason though I cannot make it catch the previous regression (i.e. duplicate expressions in group).

AlSchlo:
Notably:
```
    /// Takes the set of [`LogicalExpressionId`] that reference a group, mapped to their
    /// representatives.
    fn take_referencing_expr_set(&mut self, group_id: GroupId) -> HashSet<LogicalExpressionId> {
        self.group_referencing_exprs_index
            .remove(&group_id)
            .unwrap_or_default()
            .iter()
            .map(|id| self.repr_logical_expr_id.find(id))
            .collect()
    }

```
in helpers.rs when removing the remapping to representative logic SHOULD fail. And it does when I used to run the other testbench.

jopereira:
I'll check this today.

jopereira:
It was a typo in the test, discarding the result from shuffle method. This reproduces the bug when commenting out the remapping.

AlSchlo:
I still cannot replicate it.
```rust
/// Takes the set of [`LogicalExpressionId`] that reference a group, mapped to their
/// representatives.
fn take_referencing_expr_set(&mut self, group_id: GroupId) -> HashSet<LogicalExpressionId> {
    self.group_referencing_exprs_index
        .remove(&group_id)
        .unwrap_or_default()
}
```

Was this the change you tried out?

AlSchlo:
I believe it might be not the correct seed. However, even when trying out the old memobench, I cannot replicate it unless I use your seed (I even ran it for one hour to find another seed, but cannot).

jopereira:
Yes, that's what I'm testing and getting it to fail with pretty much any seed. Let me try with a clean checkout of this branch just to be sure.

jopereira:
It still works for me, i.e. reproduces the problem with a fresh clone of this PR. StdRng is not supposed to be fully reproducible on different systems, but it is so easy to reproduce that I find it hard to believe that it is that. Can you help me reproduce your testing environment?

PR #114, opened 2025-06-05
Add tracing support

lonless9:
## Problem

Since optd is built on top of the asynchronous runtime tokio, traditional debugging (with print statements or gdb for example) is quite difficult. What most async Rust projects use instead is logging and tracing, specifically the [tracing](https:/tracing/latest/tracing/) library.

## Summary of changes

Added tracing support for the majority of functions that need to be traced. Placed a basic subscriber in optd_cli.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2024,6 +2024,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "tracing-test",
  "trait-variant",
 ]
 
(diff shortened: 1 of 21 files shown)


lonless9:
Tracing is a quite flexible crate, and you can almost effortlessly place the spans you want anywhere. What needs careful consideration next is how to customize the subscribers built specifically for particular tasks, and change the related trace level/span/target. And these subscribers should be placed outside of optd finally.

We now need a standardized optd document template to explain how to use tracing as well as how to customize subscribers.

By the way. Once the doc is clear, we can convert all existing debug statements that use printf or eprintf to tracing under corresponding filtering rules.

